PR #24, opened 2020-01-04
using original @prefix style

josd:


domel:
Why change PREFIX to @prefix? PREFIX (SPARQL style) is valid across Turtle grammar. I suppose also with the future grammar of N3.

josd:
I prefer 1 way to express things and not 2 and I am currently switching back all my files to the original @prefix style.

domel:
@josd I understand you. But I think the priority is to create N3 grammar as superset of Turtle grammar (all ttl should be valid n3). 
Look at it like two ways of expressing bnodes. N3/Turtle you can write `_:x` and `[ ]`.

josd:
Of course, EYE also supports both as input but for the output it uses the original @prefix style. To me CWM is too important and until it supports the PREFIX style EYE will stick to the original @prefix style for all n3 that it produces. The files that are in this pull request all came from http://eulersharp.sourceforge.net/ and there we now only have the original @prefix style.

ericprud:
What if instead we update cwm?

I understand the elegance of the @`directive` design that drove `@prefix` and `@base` but I don't think it justifies the user complexity and perception of line noise. As a thought experiment, consider whether SPARQL would have appear more academic and less industrial if it used `@prefix p: <ns> .` instead of `PREFIX p: <ns>` (which is also case insensitive).

josd:
Very well Eric, I'm fine with that and will stay tuned at https://github.com/linkeddata/swap to see the new cwm ;-)

william-vw:
@josd so should we merge this pull request or let it be for now? ..

josd:
It brings the test files in line with their origin at
http://eulersharp.sourceforge.net/2007/07test/
which is when they were last produced around june 2014 like you can see in
e.g.
http://eulersharp.sourceforge.net/2007/07test/alpE003.n3 i.e.
#Processed by Id: euler.yap 7292 2014-06-13 12:27:29Z josd
A few years later I *manually* replaced the @prefix into PREFIX and that is
now reverted.

-- https://josd.github.io/ <http://josd.github.io/>

PR #31, opened 2020-07-09
EBNF grammar re-generation based on latest HTML version

gkellogg:
and replace inline HTML with data-include.

In the future, modifying grammar/n3.ebnf makes it easier to pick up changes and regenerate.


william-vw:
Something I just noticed after merging: one runs into the same-origin security restriction when doing a data-include and trying to view the page on your local file system. I actually read about this before on the [Respec Wiki](https://github.com/w3c/respec/wiki/ReSpec-Editor's-Guide#inclusions--transformations) but forgot about it. This will make it a bit more difficult to work on the spec, but not by much - one can easily run a local HTTP server (e.g., "python -m SimpleHTTPServer") on the git folder.

william-vw:
FYI: I added a README.md to the spec folder on this issue.

william-vw:
Also: how would one generate the `n3.html` from `n3.ebnf`?

gkellogg:
> Something I just noticed after merging: one runs into the same-origin security restriction when doing a data-include and trying to view the page on your local file system. I actually read about this before on the [Respec Wiki](https://github.com/w3c/respec/wiki/ReSpec-Editor's-Guide#inclusions--transformations) but forgot about it. This will make it a bit more difficult to work on the spec, but not by much - one can easily run a local HTTP server (e.g., "python -m SimpleHTTPServer") on the git folder.

I have gotten into the habit of always using SimpleHTTPServer (or python3 -m http.server) because of these issues. I just keep it going in a terminal and refresh the page when done. It seems like a small price to pay.

gkellogg:
> Also: how would one generate the `n3.html` from `n3.ebnf`?

I can include some files, but basically, you'd install the "ebnf" gem ("gem install ebnf") and run "ebnf n3.ebnf --format html -o n3.html".

However, I need to release the work on the ebnf gem to do this, so won't work now (without a Gemfile kludge to run off of GitHub).

To get it going now, you can clone https://github.com/dryruby/ebnf, run "bundler install" within that directory (presuming a minimal Ruby environment) and then run bin/ebnf from within that directory.

Once I've released an update to the gem, I can add a README to the grammar directory that shows how to do the rebuild.

william-vw:
> Once I've released an update to the gem, I can add a README to the grammar directory that shows how to do the rebuild.

That'd be great, thanks. But, I can confirm that the current process outlined above works for now!

PR #35, opened 2020-07-22
Test suite updates

gkellogg:
* Update grammar test manifests, add README and LICENSE.
* Use Turtle tests from w3c.github.io/rdf-tests/turtle/ rather than originals, as they include fixes due to errata.
* Remove negative Turtle tests an N3 processor should pass.
* Use test:TestN3Positive/NegativeSyntax instead of rdft:TestTurtlePositive/NegativeSyntax in N3 grammar tests.
* Separate extra_bad_prefix.n3 from extra_bad_prefix2.n3.

Note that all N3Tests are SyntaxTests rather than Evaluation Tests. This includes N-Triples results which were expected to be Evaluation test results. Syntax tests only how that a processor doesn't throw an error, not that it generates the right output. This is hard using Graph Isomorphism when Formulae are used, but many of the previous evaluation tests could generate N-Triples which can be successfully compared.


william-vw:
Thanks a lot for this @gkellogg. Indeed, at this stage, the tests were meant for syntax conformance more than anything else (such as appropriate interpretation of N3 documents). Essentially, they were meant to check whether the current N3 ANLTR grammar `n3.g4` could handle all the N3 constructs in these tests. 

Note that the `TurtleTests/` were really meant to check the Turtle ANTLR grammar (`turtle.g4`), which was used as a basis for the N3 ANTLR grammar. This is why these negative Turtle tests (which should indeed work for N3 processors) were listed as such in that manifest. But, this is more of an artifact of the past than anything else, i.e., a sanity check for the original basis of `n3.g4`. Maybe an option could be to have a separate "n3" and "turtle" manifest for `TurtleTests/`?

I'm a bit confused by the final N3 manifest file - does it again refer to Test[Positive/Negative]TurtleTest?

gkellogg:
> Thanks a lot for this @gkellogg. Indeed, at this stage, the tests were meant for syntax conformance more than anything else (such as appropriate interpretation of N3 documents). Essentially, they were meant to check whether the current N3 ANLTR grammar `n3.g4` could handle all the N3 constructs in these tests.
> 
> Note that the `TurtleTests/` were really meant to check the Turtle ANTLR grammar (`turtle.g4`), which was used as a basis for the N3 ANTLR grammar. This is why these negative Turtle tests (which should indeed work for N3 processors) were listed as such in that manifest. But, this is more of an artifact of the past than anything else, i.e., a sanity check for the original basis of `n3.g4`. Maybe an option could be to have a separate "n3" and "turtle" manifest for `TurtleTests/`?

The Turtle tests are, of course, maintained along with other grammar tests, in w3c.github.io/rdf-tests/turtle, so for a turtle processor, that should be sufficient. I understood our version to be for n3 processors to pass appropriate Turtle tests, which I find useful. If you want to test your Turtle Antlr processor, why should that be part of the n3 project?

> I'm a bit confused by the final N3 manifest file - does it again refer to Test[Positive/Negative]TurtleTest?

The test classes were updated to use the equivalent classes for N3. These are N3, not Turtle tests. A Turtle processor wouldn’t be able to process many of the files. Many test harnesses configure themselves based on the type, and the test vocabulary defines what these types mean for testing.

william-vw:
> The Turtle tests are, of course, maintained along with other grammar tests, in w3c.github.io/rdf-tests/turtle, so for a turtle processor, that should be sufficient. I understood our version to be for n3 processors to pass appropriate Turtle tests, which I find useful. If you want to test your Turtle Antlr processor, why should that be part of the n3 project?

> > I'm a bit confused by the final N3 manifest file - does it again refer to Test[Positive/Negative]TurtleTest?
> 
> The test classes were updated to use the equivalent classes for N3. These are N3, not Turtle tests. A Turtle processor wouldn’t be able to process many of the files. Many test harnesses configure themselves based on the type, and the test vocabulary defines what these types mean for testing.

Maybe I'm misunderstanding - but if we're tailoring the Turtle manifest file (sorry, misnamed it in my initial message) towards N3 processors, it is a bit confusing to still treat the involved tests as "Turtle tests". For instance, negative Turtle tests had to be removed from the manifest; normally these would simply be flagged as positive, but they are of course not positive as far as the "Turtle" grammar is concerned. So, the turtle manifest becomes a bit of a compromise where some tests are (rather confusingly) removed instead.

I think that having one manifest where the tests from the Turtle test suite are treated as "N3 tests" (flagging "bad" turtle tests as "positive" n3 tests) and another where they are "Turtle tests" would avoid confusion, at least for me.

Also, it would allow to more easily test and maintain the turtle grammar that was used as basis for current n3 grammar, but that is more of a secondary reason if anything.

william-vw:
anyhow this is not meant to stop from merging this pull request - we can always sort this out later.

gkellogg:
> > The Turtle tests are, of course, maintained along with other grammar tests, in w3c.github.io/rdf-tests/turtle, so for a turtle processor, that should be sufficient. I understood our version to be for n3 processors to pass appropriate Turtle tests, which I find useful. If you want to test your Turtle Antlr processor, why should that be part of the n3 project?
> 
> > > I'm a bit confused by the final N3 manifest file - does it again refer to Test[Positive/Negative]TurtleTest?
> > 
> > 
> > The test classes were updated to use the equivalent classes for N3. These are N3, not Turtle tests. A Turtle processor wouldn’t be able to process many of the files. Many test harnesses configure themselves based on the type, and the test vocabulary defines what these types mean for testing.
> 
> Maybe I'm misunderstanding - but if we're tailoring the Turtle manifest file (sorry, misnamed it in my initial message) towards N3 processors, it is a bit confusing to still treat the involved tests as "Turtle tests". For instance, negative Turtle tests had to be removed from the manifest; normally these would simply be flagged as positive, but they are of course not positive as far as the "Turtle" grammar is concerned. So, the turtle manifest becomes a bit of a compromise where some tests are (rather confusingly) removed instead.

I'd be fine with re-typing them as `TestN3PositiveSyntax` or equivalent and making the former negative tests positive.

> I think that having one manifest where the tests from the Turtle test suite are treated as "N3 tests" (flagging "bad" turtle tests as "positive" n3 tests) and another where they are "Turtle tests" would avoid confusion, at least for me.
> 
> Also, it would allow to more easily test and maintain the turtle grammar that was used as basis for current n3 grammar, but that is more of a secondary reason if anything.

So, you'd like to see the original manifest.ttl with all Turtle tests, and a new manifest.n3 with the N3 versions? Still, I don't understand what that buys you that just targeting https://w3c.github.io/rdf-tests/turtle doesn't already do. While there is some possibility for ongoing maintenance and divergence with the Turtle test suite, it's probably pretty minor until that spec is revisited.

I will have additional PRs to clean up some test expectation, and updates where we've changed the grammar (e.g., `@false`). Let's clarify exactly what the purpose of the test suites are and how they should change going forward.

william-vw:
> So, you'd like to see the original manifest.ttl with all Turtle tests, and a new manifest.n3 with the N3 versions? Still, I don't understand what that buys you that just targeting https://w3c.github.io/rdf-tests/turtle doesn't already do. While there is some possibility for ongoing maintenance and divergence with the Turtle test suite, it's probably pretty minor until that spec is revisited.

Mostly it'd just make it slightly easier to run the Turtle tests really, nothing more. But if we add our own `manifest.ttl` file, i.e., which annotates the tests according to their N3 conformance, we'd at least know that this manifest file is up-to-date with our current snapshot of Turtle tests (unless the Turtle people want to keep it up to date as well). But that's more of a minor issue really.

gkellogg:
I can see changing the types in the turtle manifest to `test:TestN3Eval`, instead of `rdft:TestTurtleEval`, and so forth. We could presumably add both the Turtle and N3 types to each manifest entry, accounting for variations in expectations.

Given that the Turtle suite is curated by the RDF Testing CG (for which I happen to be the chair), I could also see adding these to the Turtle manifest in w3c/rdf-tests, which I presume would not have any objections for a PR. We could also integrate the N3-specific test vocabulary into `rdft: <http://www.w3.org/ns/rdftest#>`, which would make sense if we do the second approach.

The RDF Tests repository is live and taking updates the community deems fit.

william-vw:
Yes, I think having separate N3 test entries within the same manifest file (with different expected outcomes in some cases) would be a good solution.

Moving forward, having a [https://w3c.github.io/rdf-tests/n3](https://w3c.github.io/rdf-tests/n3) with N3-specific tests would be really neat. A "master" manifest file (such as the one you've added under `grammar/tests/` referring to several other manifests) could then reference the N3-specific manifest and the Turtle manifest (with N3 entries) within that repo. 

But, to your point, we'll have to properly structure the N3 manifest since it's focusing solely on syntax currently (I think that was done properly in your N3 test manifest?).

gkellogg:
> Yes, I think having separate N3 test entries within the same manifest file (with different expected outcomes in some cases) would be a good solution.

I'll take care of that, and do some other house-keeping.

> Moving forward, having a https://w3c.github.io/rdf-tests/n3 with N3-specific tests would be really neat. A "master" manifest file (such as the one you've added under `grammar/tests/` referring to several other manifests) could then reference the N3-specific manifest and the Turtle manifest (with N3 entries) within that repo.

No reason it can't reference a remote context, too.

> But, to your point, we'll have to properly structure the N3 manifest since it's focusing solely on syntax currently (I think that was done properly in your N3 test manifest?).

Yes, and I'll look to either make positive syntax tests evaluation tests, or move over some of the others and generally clean up the /tests directory.

PR #36, opened 2020-07-24
More Test Suite updates

gkellogg:
* Update manifest to reorder test sequence and restore evaluation tests from old manifest-grammar.
* Remove `rdft:approval` from syntax tests, which certainly haven't been approved yet.
* Add N3 test classes to Turtle tests.


william-vw:
Thanks for looking into these tests @gkellogg!

In general, the following files are a bit confusing:
- `turtle-syntax-bad-n3-extras-03.ttl`: comment referring to "N3 paths" but the code does not contain an N3 path
- `turtle-syntax-bad-n3-extras-06.ttl`: comment saying "=" is not in Turtle but the code doesn't contain that symbol 

Both are actually parsed by the N3 parser; the first one since since statements with only a subject are allowed, and the second one since `.` are allowed in `PN_LOCAL` (both in N3 and Turtle). The Turtle parser throws an error fr both since statements with only a subject are not accepted. Try them in the N3 editor [here](http://ppr.cs.dal.ca:3002/n3/editor/?formula=%40prefix%20%3A%20%3Chttp%3A%2F%2Fwww.w3.org%2F2013%2FTurtleTests%2F%3E%20.%0A%40prefix%20ns%3A%20%3Chttp%3A%2F%2Fwww.w3.org%2F2013%2FTurtleTests%2Fp%23%3E%20.%0A%0A%3Ax.%0A%20%20ns%3Ap.%0A%20%20%20%20ns%3Aq%20%3Ap%20%3Az%20.%0A) and [here](http://ppr.cs.dal.ca:3002/n3/editor/?formula=%40prefix%20%3A%20%3Chttp%3A%2F%2Fwww.w3.org%2F2013%2FTurtleTests%2F%3E%20.%0A%0A%3Aa.%3Ab.%3Ac%20.%0A) (try showing the AST or executing with eye / cwm)

gkellogg:
turtle-syntax-bad-n3-extras-03.ttl contains the following:

```turtle
:x.
  ns:p.
    ns:q :p :z .
```

where `ns:q :p :z` is a path.

For turtle-syntax-bad-n3-extras-06.ttl, I'll update the comment (and eventually sync back to rdf-tests); it uses the old "." notation paths, which is invalid now in N3, as well:

```turtle
:a.:b.:c .
```

gkellogg:
It's turtle-syntax-bad-n3-extras-02 that has "= is not Turtle (negative test)", and it does contain an `=`:

```turtle
:a = :b .
```

I wouldn't be surprised to find more where the comments are at odds with the actual tests.

gkellogg:
One thing I'd like to do is to either add some option annotations, or split the N3 syntax tests so that those that take particularly long to run (at least for me) can be separated from those that run reasonably quickly to aid in development cycles.

Traditionally, tests are narrowly focused on the specific feature being tested, but having stress tests is useful, at least for implementors if not to prove conformance.

william-vw:
> 
> 
> turtle-syntax-bad-n3-extras-03.ttl contains the following:
> 
> ```turtle
> :x.
>   ns:p.
>     ns:q :p :z .
> ```
> 
> where `ns:q :p :z` is a path.

I'm unaware of this old "path" syntax but it seems that `ns:q :p :z` is simply a statement, no?

> For turtle-syntax-bad-n3-extras-06.ttl, I'll update the comment (and eventually sync back to rdf-tests); it uses the old "." notation paths, which is invalid now in N3, as well:
> 
> ```turtle
> :a.:b.:c .
> ```

Note that both syntax tests work in N3 (perhaps not clear from my message), so I don't think they should be *negative* N3 syntax tests. But, there could be eval tests for checking whether the right stuff is being parsed from these files (i.e., not N3 paths, but what I described in the prior message).

william-vw:
> 
> 
> One thing I'd like to do is to either add some option annotations, or split the N3 syntax tests so that those that take particularly long to run (at least for me) can be separated from those that run reasonably quickly to aid in development cycles.
> 
> Traditionally, tests are narrowly focused on the specific feature being tested, but having stress tests is useful, at least for implementors if not to prove conformance.

Yes, good idea. There's some files in there that take a long time to parse (e.g., "data.n3").

gkellogg:
> Note that both syntax tests work in N3 (perhaps not clear from my message), so I don't think they should be _negative_ N3 syntax tests.

I don't think they're valid.

```
:x.
  ns:p.
    ns:q :p :z .
```

Was intended to be the same as `:x!ns:p!ns:q :p :z` in the original syntax. Parsed as such, it might look like three different statements, but the first two would have just a subject, and no predicate or object, so would not be valid from my interpretation of the grammar.

```
:a.:b.:c .
```

Could be interpreted as `:a!:b!:c .` using the old grammar, but as `:a.` isn't an IRI, it wouldn't parse either, so they seem like negative tests to me.

william-vw:
> 
> 
> > Note that both syntax tests work in N3 (perhaps not clear from my message), so I don't think they should be _negative_ N3 syntax tests.
> 
> I don't think they're valid.
> 
> ```
> :x.
>   ns:p.
>     ns:q :p :z .
> ```
> 
> Was intended to be the same as `:x!ns:p!ns:q :p :z` in the original syntax. Parsed as such, it might look like three different statements, but the first two would have just a subject, and no predicate or object, so would not be valid from my interpretation of the grammar.

The prior (and current) grammar perfectly allows statements with only a subject. See [triples](https://w3c.github.io/N3/spec/#grammar-production-triples) production.
> 
> ```
> :a.:b.:c .
> ```
> 
> Could be interpreted as `:a!:b!:c .` using the old grammar, but as `:a.` isn't an IRI, it wouldn't parse either, so they seem like negative tests to me.

As said, `.` are explicitly allowed in [PN_LOCAL ](https://w3c.github.io/N3/spec/#grammar-production-PN_LOCAL)(both in N3 and Turtle) so there's no reason that prefixed name wouldn't parse.

gkellogg:
Ahh, yes, a parser update missed that. I'll rework and do another test update.

william-vw:
Just let me know when you think it's ready to be merged (perhaps easier to accumulate that into a single PR)

(e.g., I don't think the `..n3-extras-03.ttl` and `..n3-extras-06.ttl` tests are correctly annotated at this point)

gkellogg:
I continue to go through tests, and haven't spent too much time on the extended tests.

As for major structural changes, I think we should move grammar/tests/ into /tests, and reconcile the reasoner tests with the newer naming convention. If you agree with that, I'll work on that next, then we should be good to go.

gkellogg:
That should be enough changes for now. It should be good to go.

william-vw:
> 
> 
> I continue to go through tests, and haven't spent too much time on the extended tests.
> 
> As for major structural changes, I think we should move grammar/tests/ into /tests, and reconcile the reasoner tests with the newer naming convention. If you agree with that, I'll work on that next, then we should be good to go.

Yes, I agree - thanks for this.

Issue #48, opened 2020-08-27
string:concatenation of non-string values


gkellogg:
string:concatenation is defined to work on string values, but is often used for other literal values. For example, the following tests:

[cwm_includes/concat.n3]()

    {
      ("CASH"^^xsd:string
       " IS "^^xsd:Name
       ("2"^^xsd:integer "3"^^xsd:unsignedShort)!math:product
       "\n"
      ) string:concatenation  :s}
    => {"Here we go"^^xsd:string log:outputString :s} .

[cwm_math/math-test.n3]()

    "3.1415926" a :testValue.
    3.1415926 a :testValue.
    "1729" a :testValue.
    1729 a :testValue.
    "0" a :testValue.
    0 a :testValue.
    "1.0e7" a :testValue.
    1.0e7 a :testValue.

    { :x a :testValue. :y a :testValue.
      (:x :y) math:sum :z.
      (:x " + " :y " = " :z ) string:concatenation ?s
    } => { ?s a :RESULT }.

The test results often assume some native transformation of these values, which is non-standard. I'd suggest that any values turned into strings be canonicalized using the forms defined for the [XSD canonical representation](https://www.w3.org/TR/xmlschema11-2/#dt-canonical-representation) of datatyped literals, although IRIs would also be expected in canonical form, for consistency.

Note, for double (for which XSD is vague) it would typically be represented by `"\-?\d+\.\d+E\-?\d+"` with no trailing or leading zeros.

In a related matter, considering how values are compared in tests, a parsed INTEGER,DECIMAL,DOUBLE should probably be represented in canonical form as well, so that calculations can compare to the literal defined in the test result, or at least, this should be mentioned in considerations when running tests.

gkellogg:
For backwards compatibility, we probably need to break with other RDF forms and stick with a lower-case "e" for doubles, otherwise, revisit a lot of test results.

josd:
All I can say at this moment is that for try.n3 being
```
@prefix math: <http://www.w3.org/2000/10/swap/math#>.
@prefix string: <http://www.w3.org/2000/10/swap/string#>.
@prefix xsd: <http://www.w3.org/2001/XMLSchema#>.
@prefix : <http://example.org/test#>.

{   (   "CASH"^^xsd:string
        " IS "^^xsd:Name
        ("2"^^xsd:integer "3"^^xsd:unsignedShort)!math:product
        "\n"
    ) string:concatenation ?S.
} => {
    ?S a :OK.
}.

[...]

Issue #46, opened 2020-08-27
value space of math:exponentiation


gkellogg:
From the tests, it seems that `(2 10) math:exponentiation ?x` results in an integer, but [XPath defines math:pow](https://www.w3.org/TR/xpath-functions-31/#func-math-pow) to result in a double:

> math:pow($x as xs:double?, $y as xs:numeric) as xs:double?

Some cases it may result in an integer, but others a double, or maybe decimal. To do anything other than a double would seem to require some heuristics on the input. How do we want to define this?

Some possibilities:

* integer ^ nonNegativeInteger => integer
* integer ^ other numeric => double
* decimal ^ nonNegativeInteger => decimal
* decimal ^ other numeric => double
* double ^ any numeric => double

What do EYE and CWM do?

In keeping with SPARQL functions, we may want to standardize on the XPath `math:pow` definition.

josd:
Both EYE and CWM   `(2 10) math:exponentiation 1024`
Here are some more results for EYE:
```
(2 10.3) math:exponentiation 1260.691879265195 .
(2.7 10) math:exponentiation 20589.113209464915 .
(2.7 10.3) math:exponentiation 27736.18790208095 .
(2.7e+20 10.3) math:exponentiation 2.773618790208184e+210 .
(200.0 103.0) math:exponentiation 1.0141204801825836e+237 .
```
Using the e:derive built-in EYE can also do complex number exponentiation such as the
Euler identity `("exponentiation" ("e" 0) (0 "pi") (-1.0 1.2246467991473532e-16)) e:derive true.`
https://github.com/josd/eye/blob/master/reasoning/complex/complex_query.n3
https://github.com/josd/eye/blob/master/reasoning/complex/complex_answer.n3

josd:
To come back to the last example, it really make sense to have integers where the result is an integer:
`(200 103) math:exponentiation 1014120480182583521197362564300800000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000 .`

josd:
It is actually very similar to math:sum
```
(1 1) math:sum 2 .
(1.0 1) math:sum 2.0 .
```
and there is no reason to make (1 1)!math:sum a xsd:double

gkellogg:
[...]

PR #50, opened 2020-08-29
More Reasoner tests

gkellogg:
* Move newly created tests to their own directories, not cwm_ directories.
* Tests for `list:first`, and `list:length`.
* Tests for `math:absoluteValue`, `math:ceiling`, and `math:floor`.
* Version of `math-test-ref.n3` added, although concerns over results (see #46 and #48).
* Add tests for `math:exponentiation`, `math:difference`, `math:product`, `math:quotient`, `math:remainder`, and `math:sum` (also relates to #48 discussion about the promotion or demotion of numeric types.).

@cc/ @josd


gkellogg:
The math test results differ from what I believe cwm and eye do with regards to certain math operations on mixed numeric types, but are what I think are correct. It could be that different platforms will give slightly different values for floating point operations, so we may need to choose operands which are simpler to process for different math packages.

Issue #51, opened 2020-09-01
Time builtins


gkellogg:
## Multiple vocabularies

There apparently are two different time vocabularies, [time.n3](http://www.w3.org/2000/10/swap/time.n3), and [times.n3](http://www.w3.org/2000/10/swap/times.n3), although times.n3 declares vocabulary terms in the "time" namespace. They both describe resulting results as strings, although the [t1.n3](https://github.com/w3c/N3/blob/master/tests/N3Tests/cwm_time/t1.n3) tests expect numbers, and the [t1s.n3](https://github.com/w3c/N3/blob/master/tests/N3Tests/cwm_time/t1s.n3) expect string results. We should pick one or the other behavior.

## time:dayOfWeek wrong.

The expected results for "1999-12-31T23:59:59.99Z" is `4` (where Sunday is 0), which would mean that 31 December 1999 fell on a Thursday, when in fact, [it is on a Friday](https://www.calendar-365.com/calendar/1999/December.html). The CWM python code for this seems to be doing a shortcut:

```python
class BI_dayOfWeek(LightBuiltIn, Function):
    def evaluateObject(self,  subj_py):
        weekdayZero = time.gmtime(0)[6]  // evaluates to 3
        return (weekdayZero + int(isodate.parse(subj_py)/DAY)) % 7 
```

My own code uses Ruby's `strftime` package, which is pretty universal:

```ruby
resource = DateTime.iso8601(subject.to_s)
RDF::Literal(resource.object.strftime("%w").to_i)
```

This gives the correct answer of 5.

I propose to update the expected results accordingly.

gkellogg:
I believe we've resolved this and can close.

Issue #47, opened 2020-08-27
Numeric interpretation of plain literals


gkellogg:
There are a number of cases where the math functions operate on plain literals (e.g., `{ ("3" "5") math:sum :x } => { :x :valueOf "3 + 5" } .`. Typically, these have the form of integers, but that might not always be the case. Do we define how to interpret plain literals as numbers? We could use the `numericLiteral` production to intuit this, which has precedent in `log:semantics`. We should define it, though.

gkellogg:
Similarly, we should define how a processor should behave when a literal does not parse as any of double, decimal, or integer, or if just a prefix of the literal parses that way.

gkellogg:
I could see something like a `math:cast` operator (like SPARQL STRDT) doing something like the following:

`{("1" xsd:integer) log:STRDT ?x} => {:result :is ?x} .`

resulting in `:result :is 1 .`.

It would also be useful to be able to extract the datatype or language or a literal
```turtle
{ 1.1 log:datatype ?y } => {1.1 :datatype ?y } .
```

resulting in `1.1 :datatype xsd:decimal .`.

Other functions from [SPARQL Functions on RDF Terms](https://www.w3.org/TR/sparql11-query/#func-rdfTerms) would also be useful:

* `isIRI`
* `isBlank`
* `isLiteral`
* `isNumeric`
* `str`
* `lang`
* `datatype`
* `IRI`
* `BNODE`
* `STRDT`
* `STRLANG`

They might even be bi-directional.

pchampin:
:+1: for `math:cast`.

---

However, I am not in favour of `isIRI`, `isBlank`, `isLiteral`. These SPARQL functions apply to the *syntax*, not the semantics. So stating something like
```turtle
<http://champin.net/#pa> log:isIRI true.
```
violates the [use/mention distinction](https://en.wikipedia.org/wiki/Use%E2%80%93mention_distinction). “I am not an IRI, I'm a free man!” :stuck_out_tongue_winking_eye: 
Put it another way, what would you expect from the following?
```turtle
:answer = 42.
{ :answer log:isLiteral ?x } => { :test1 :gave ?x }.
{ 42 log:isIRI ?x } => { :test2 :gave ?x }.
```
---

`log:datatype` may raises a similar problem (depending how we consider it).
What would you expect from the following?
```turtle
{ "42"^^xsd:integer log:datatype ?x } => { :test3 :gave ?x }.
```
[...]

PR #54, opened 2020-09-09
Numeric and date/time tests.

gkellogg:
Note, this updates some test expectation to use canonical (more or less) representations of numeric values. For double, this uses, e.g., `1.0e0`. Strictly speaking, according to XSD, this should be `1.0E0`, but this would have even wider implications for existing implementations, we should discuss. Also, it does not reflect interpretation of plain literals #47.

Both forms of time tests (string results and numeric results) are used. this relates to #51.

Per @pchampin's suggestion, I've added a number of tests for infinities and NaN. We might also consider adding `[+-]INF` and `NaN` to the syntax.


pchampin:
> We might also consider adding [+-]INF and NaN to the syntax.

I'm not too sure about it. This would add complexity to the parsers (think `:a :b INF, INF:c`) for what I consider to be a niche use case. And this does not make the language more expressive, as these values *can* be expressed with the `"..."^^<...>` syntax already.

PR #58, opened 2020-09-28
Namespace imports

gkellogg:
Imports the SWAP versions of the builtin namespace documents, with updates.

See notes in the commit for further discussion.

### DIFF
--- a/ns/crypto.n3
+++ b/ns/crypto.n3
@@ -1,0 +1,179 @@
+#  Schema for crypto built-in functions in cwm. 
+#
+#  These are terms drawn to match the available functions
+# in mxCrypto. See:
+# http://www.amk.ca/python/writing/pycrypt/  for the manual and
+# http://www.amk.ca/python/code/crypto.html  for the code.
+#
+#  I am not sure whether in the domain and range functions we should
+# distinguish strings one can present to a person from crypto binary.
+# Might be useful for driving the UI and/or serialization using base64 etc.
+#
+#  Beware that the simplicity of use of some of these properties belies
+# the knowledge o fcrypography that one needs to make a secure system
+#
+@prefix rdf:  <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
+@prefix s: <http://www.w3.org/2000/01/rdf-schema#> .
+@prefix owl: <http://www.w3.org/2002/07/owl#>.
+# @prefix daml:  <http://www.daml.org/2001/03/daml-ont#> .
+# @prefix dpo:  <http://www.daml.org/2001/03/daml+oil#> .
+@prefix log:  <http://www.w3.org/2000/10/swap/log#> .
+
+@prefix string:  <http://www.w3.org/2000/10/swap/string#> .
+@prefix cr:  <http://www.w3.org/2000/10/swap/crypto#> .
+
+
+#####################################################################
+#
+# Hash functions
+
+cr:HashFunction a s:Class;
+    s:label "Any crypographic hash function";
+    s:comment """
+The crypographic hash functions are (being functions) unique and are,
+when secure, assumed unambiguous (the whole point of being hash
+functions). That is, when you have the right hash, you have the right
+document. Currently (2001/9) only SHA is given that property.
+""".
+
+#cr:HAVAL128 a rdf:Property; s:label "HAVAL hash (128 bit)";
+#    s:comment "The object is a HAVAL hash (128 bit) of the subject.";
+#    a dpo:UniqueProperty;
+#    s:domain string:String;
+#    s:range string:String.
+#
+#cr:HAVAL160 a rdf:Property; s:label "HAVAL hash (160 bit)";
+#    s:comment "The object is a HAVAL hash (160 bit) of the subject.";
+#    a dpo:UniqueProperty;
+#    s:domain string:String;
+#    s:range string:String.
+#
+#cr:HAVAL192 a rdf:Property; s:label "HAVAL hash (192 bit)";
+#    s:comment "The object is a HAVAL hash (192 bit) of the subject.";
+#    a dpo:UniqueProperty;
+#    s:domain string:String;
+#    s:range string:String.
+#
+#cr:HAVAL224 a rdf:Property; s:label "HAVAL hash (224 bit)";
+#    s:comment "The object is a HAVAL hash (224 bit) of the subject.";
+#    a dpo:UniqueProperty;
+#    s:domain string:String;
+#    s:range string:String.
+#
+#cr:HAVAL256 a rdf:Property; s:label "HAVAL hash (256 bit)";
+#    s:comment "The object is a HAVAL hash (256 bit) of the subject.";
+#    a dpo:UniqueProperty;
+#    s:domain string:String;
+#    s:range string:String.
+#
+#cr:MD2 a rdf:Property; s:label "MD2 hash";
+#    s:comment "The object is a MD2 hash of the subject.";
+#    a dpo:UniqueProperty;
+#    s:domain string:String;
+#    s:range string:String.
+#
+#cr:MD4 a rdf:Property; s:label "MD4 hash";
+#    s:comment "The object is a MD4 hash of the subject.";
+#    a dpo:UniqueProperty;
+#    s:domain string:String;
+#    s:range string:String.
+#
+cr:md5 a rdf:Property; s:label "MD5 hash";
+    s:comment "The object is a MD5 hash of the subject.";
+    a owl:FunctionalProperty;
+    s:domain string:String;
+    s:range string:String.
+
+#cr:RIPEMD a rdf:Property; s:label "RIPEMD hash";
+#    s:comment "The object is a RIPEMD hash of the subject.";
+#    a dpo:UniqueProperty;
+#    s:domain string:String;
+#    s:range string:String.
+
+cr:sha a rdf:Property; s:label "SHA hash";
+    s:comment "The object is a SHA-1 hash of the subject.";
+    a owl:FunctionalProperty, owl:InverseFunctionalProperty;  # Assume unbreakable.
+    s:domain string:String;
+    s:range string:String.
+
+###########################  Public Key Algorithms
+#
+# See http://www.amk.ca/python/writing/pycrypt/node16.html
+#
+
+cr:PublicKeyObject a s:Class;
+    s:comment """An object corresponding to a key for some algorithm.
+The object can hold a public and optionally a private key.""".
+
+cr:CanEncrypt s:isSubClassOf cr:PublicKeyObject;
+    s:comment """PublicKeyObjects which are capable of encrypting things""".
+
+cr:CanSign s:isSubClassOf cr:PublicKeyObject;
+    s:comment """PublicKeyObjects which are capable of signing things.
+True if the algorithm is capable of signing data; false otherwise. To
+test if a given key object can sign data, use CanSign and HasPrivate.""".
+
+cr:HasPrivate s:isSubClassOf cr:PublicKeyObject;
+    s:comment"""
+Some keys have private parts, some don't.  This is the class of those which do.
+""".
+
+cr:publicKey a s:Property;
+    s:domain cr:PublicKeyObject;
+    s:range cr:PublicKeyObject;
+    s:comment """The object is a public key object that doesn't
+    contain the private key data in the subject.
+    This function extracts the public part.""".
+
+
+#cr:size a s:Property;
+#    s:domain cr:PublicKeyObject;
+#    s:range  log:Number;   # @@  string, decimal
+#    s:comment """The maximum size of a string that can be encrypted or signed,
+#measured in bits. String data is treated in big-endian format; the most significant
+#byte comes first. (This seems to be a de facto standard for cryptographical
+#software.) If the size is not a multiple of 8, then some of the high order
+#bits of the first byte must be zero. Usually it's simplest to just divide the
+#size by 8 and round down""".
+#
+#
+#cr:RSA s:SubClassOf cr:PublicKeyObject;
+#    s:comment "An RSA public key object".
+#
+#cr:RSA_ a rdf:Property;
+#    a dpo:UniqueProperty;
+#    s:label "RSA key generator";
+#    s:domain string:String;    # Actually a number in dcimal, length of ky in octets
+#    s:range  string:String.    # The key generated
+#
+#cr:RSA_validSignature a rdf:Property;   # Must be an internal built-in
+#    a dpo:UnambiguousProperty, dpo:UniqueProperty;  # Assume unbreakable.
+#    s:label "validated RSA signature";
+#    s:domain string:String;   # A signed string, typically the hash of the document contents
+#    s:range string:String.   # The signature
+
+cr:sign a rdf:Property;
+    s:label "sign";
+    s:comment """The subject should be a list of two things, a hash
+string and a key (containing private and public parts). The object is
+calculated as a signature string by signing the hash with the key's
+private part.""".
+
+cr:verify a rdf:Property;
+    s:comment """If the subject is a key object containing private and
+public parts and the obejct is a list of a hash and a signature, then
+this is true if and only if the signature is a valid signature of the
+hash with the key.""".
+
+cr:verifyBoolean a rdf:Property;
+    s:comment """If the subject is a list containg a keypair, a hash,
+and a signature, then the object is either "1" if the signature
+validates or "0" if it does not.""".
+
+####### Metaproperties
+#
+# These generate, given an encryption module, the funmctions for that module.
+#
+
+
+#ends
--- a/ns/crypto.n3
+++ b/ns/crypto.n3
@@ -13,10 +13,8 @@
 # the knowledge o fcrypography that one needs to make a secure system
 #
 @prefix rdf:  <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
-@prefix s: <http://www.w3.org/2000/01/rdf-schema#> .
+@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .
 @prefix owl: <http://www.w3.org/2002/07/owl#>.
-# @prefix daml:  <http://www.daml.org/2001/03/daml-ont#> .
-# @prefix dpo:  <http://www.daml.org/2001/03/daml+oil#> .
 @prefix log:  <http://www.w3.org/2000/10/swap/log#> .
 
 @prefix string:  <http://www.w3.org/2000/10/swap/string#> .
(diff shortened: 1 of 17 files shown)


ns/crypto.n3
```diff
@@ -0,0 +1,94 @@
...
+# Hash functions
+
+cr:HashFunction a rdfs:Class;
+    rdfs:label "Any crypographic hash function";
+    rdfs:comment """
+The crypographic hash functions are (being functions) unique and are,
+when secure, assumed unambiguous (the whole point of being hash
+functions). That is, when you have the right hash, you have the right
+document. Currently (2001/9) only SHA is given that property.
+""".
```
gkellogg, line 35: We haven't discussed this.

ns/crypto.n3
```diff
@@ -0,0 +1,94 @@
...
+
+cr:md5 a rdf:Property, owl:FunctionalProperty, owl:InverseFunctionalProperty;
+    rdfs:label "MD5 hash";
+    rdfs:comment "The object is a MD5 hash of the subject.";
+    rdfs:domain string:String;
+    rdfs:range string:String.
+
```
gkellogg, line 42: We haven't specifically described md5 hash.

ns/crypto.n3
```diff
@@ -0,0 +1,94 @@
...
+cr:sha a rdf:Property, owl:FunctionalProperty, owl:InverseFunctionalProperty;
+    rdfs:label "SHA hash";
+    rdfs:comment "The object is a SHA-1 hash of the subject.";
+    rdfs:domain string:String;
+    rdfs:range string:String.
```
gkellogg, line 47: As SHA-1 is fairly archaic, we should also consider SHA-256 and SHA-512.

ns/crypto.n3
```diff
@@ -0,0 +1,94 @@
...
+
+###########################  Public Key Algorithms
```
gkellogg, line 49: Do we use these at all?

ns/list.n3
```diff
@@ -0,0 +1,98 @@
...
+    rdfs:isDefinedBy <>.
+
+list:first a rdf:Property, owl:FunctionalProperty;
+    rdfs:label "first";
+    rdfs:domain log:List;
+    rdfs:comment """
+Iff the subject is a list and the object is the first thing that list, then this is true.
+The object can be calculated as a function of the list.
+""";
+    vs:term_status "unstable";
```
gkellogg, line 34: Note that here, and elsewhere, additions are marked with an "unstable" term status.

ns/list.n3
```diff
@@ -0,0 +1,98 @@
...
+    rdfs:label "remove";
+    rdfs:domain log:List;
+    rdfs:range log:List;
+    rdfs:comment """
+Iff the subject is a list of two lists and the object is a list composed of the
+members of the first component of the subject up to but not including the second component
+of the subject, then this is true.
+eg ( (1 2 3 4) (3 4) ) list:remove (1 2).
+The object can be calculated as a function of the subject.
+""";
```
gkellogg, line 95: This was my interpretation of the comment on how this would work, but I'm not sure it's too useful.

ns/log.n3
```diff
@@ -0,0 +1,396 @@
...
+log:List, log:Set or log:Other. 
+Example: log:semanticsOrError returns either a formula or a string,
+and you can check which using log:rawType.
+""".
+
+log:Type a rdfs:Class
+    rdfs:comment """
+The class of Notation3-specifiec resource types.
+""";
+    vs:term_status "unstable".
```
gkellogg, line 59: I created this to allow log:rawType to use it as a a range.

ns/log.n3
```diff
@@ -0,0 +1,396 @@
...
+
+log:Formula a rdfs:Class; rdfs:subClassOf log:Type.
+log:List    a rdfs:Class; rdfs:subClassOf log:Type, rdf:List.
+log:String  a rdfs:Class; rdfs:subClassOf log:Type.      # Replace with xsd:string or rdfs:Literal?
```
gkellogg, line 63: Should this just be rdfs:Literal or xsd:string? Do we need some class which specifically deals with type conversion?

ns/log.n3
```diff
@@ -0,0 +1,396 @@
...
+log:Other   a rdfs:Class; rdfs:subClassOf log:Type.     # Not one of the above.
+
+#   The following are inherent in the translation of N3 syntax to quad representation:
+
+# FIXME: no such thing as rdf:SyntaxArc.
+log:forSome a rdf:SyntaxArc;  #@@@@@   NOT a real property. A pseudo property
```
gkellogg, line 69: Note, there is no `rdf:SyntaxArc` class.

Both of these were in place, and may not bear a good relationship to what is suggested in our builtin wishlist.

ns/log.n3
```diff
@@ -0,0 +1,396 @@
...
+    rdfs:isDefinedBy <>.
+
+
+# The following connect the inference engine to the web
+
+
+log:parsedAsN3 a rdf:Property;
+    rdfs:comment "The subject string, parsed as N3, gives this formula.";
+    rdfs:label "parses to";
+    rdfs:domain log:String;
```
gkellogg, line 158: How does this relate to `log:N3Document`?

ns/log.n3
```diff
@@ -0,0 +1,396 @@
...
+    rdfs:range log:Formula.
+ 
+ log:n3String a rdf:Property;
+    rdfs:comment "The subject formula, expressed as N3, gives this string.";
+    rdfs:label "n3 string";
+    rdfs:domain log:Formula;
+    rdfs:range log:String.
```
gkellogg, line 165: How does this relate to `log:N3Document`?

ns/log.n3
```diff
@@ -0,0 +1,396 @@
...
+    rdfs:domain doc:Work;
+    rdfs:range  log:Formula;
+    rdfs:label  "semantics";
+    rdfs:comment """The log:semantics of a document is the formula.
+achieved by parsing representation of the document.
+For a document in Notation3, log:semantics is the
+log:parsedAsN3 of the log:content of the document.
+For a document in another RDF syntax, it is parsed according to the
+appropriate specification to yield an RDF formula
+(a subclass of N3 log:Formula).
```
gkellogg, line 177: Updated to be generic about other RDF syntaxes, not just RDF/XML.

ns/log.n3
```diff
@@ -0,0 +1,396 @@
...
+""" When document D is the definitiveDocument for property P,
+any statement X P Y is true iff the semantics of document D
+include that statement.
+
+For example, there may be a definitive document for the zipcode of
+airports by airport code, and so on. This is useful to let a reasoner
+know that it can extend its query to the given document.
+(Cwm will do this if its mode includes "r").
+""";
+    vs:term_status "deprecated".
```
gkellogg, line 248: Note term is marked as deprecated.

ns/log.n3
```diff
@@ -0,0 +1,396 @@
...
+    rdfs:label "output string";
+    rdfs:range log:String;
+    rdfs:comment """The subject is a key and the object is a string,
+where the strings are to be output in the order of the keys.
+See cwm --strings in cwm --help.
+""";
+    vs:term_status "deprecated".
+
+
+################### Experimental
```
gkellogg, line 282: Definitions of the following likely need more specificity.

ns/log.n3
```diff
@@ -0,0 +1,396 @@
...
+
+log:skolem a rdf:Property;
+    rdfs:label "skolem";
+    rdfs:domain log:List;
+    rdfs:range log:String;
+    rdfs:comment """
+Built-in to generate a Skolem IRI object which is a function of the arguments in the subject list""";
+    vs:term_status "unstable".
+
+log:N3 a rdfs:Datatype;
```
gkellogg, line 345: Now does this relate to `log:N3Document`?

ns/math.n3
```diff
@@ -0,0 +1,239 @@
...
+@prefix log: <http://www.w3.org/2000/10/swap/log#> .
+@prefix math: <http://www.w3.org/2000/10/swap/math#> .
+@prefix string: <http://www.w3.org/2000/10/swap/string#> .
+@prefix vs: <http://www.w3.org/2003/06/sw-vocab-status/ns#> .
+
+math:StrictProperty a rdfs:Class;
+    rdfs:label "StrictProperty";
+    rdfs:subClassOf rdf:Property.
+
+math:Value a rdfs:Class; rdfs:label "Value";
```
gkellogg, line 18: This should probably be `rdfs:subClassOf xsd:anyAtomicType`, or be specific about `xsd:decimal`, `xsd:float`, `xsd:integer`, `xsd:double`, and `xsd:string`.

ns/math.n3
```diff
@@ -0,0 +1,239 @@
...
+
+math:StrictProperty a rdfs:Class;
+    rdfs:label "StrictProperty";
+    rdfs:subClassOf rdf:Property.
+
+math:Value a rdfs:Class; rdfs:label "Value"; 
+   rdfs:subClassOf string:String; 
+   rdfs:comment "The class of things that are numeric float values as in Python." .
+
+math:Function a rdfs:Class;
```
gkellogg, line 22: Should use of this be changed to `owl:FunctionalProperty`?

ns/math.n3
```diff
@@ -0,0 +1,239 @@
...
+   rdfs:subClassOf rdf:Property;
+   rdfs:label "Function";
+   rdfs:comment "A math:Function is unique in terms of math:EqualTo. ".
+
+math:ReverseFunction a rdfs:Class;
```
gkellogg, line 27: Should use of this be changed to `owl:InverseFunctionalProperty`?

ns/time.n3
```diff
@@ -0,0 +1,173 @@
...
+
+@prefix time: <http://www.w3.org/2000/10/swap/time#>.
+@prefix : <#>.
+
+<> <http://www.w3.org/2000/10/swap/pim/doc.n3#licence> <http://www.w3.org/Consortium/Legal/copyright-software-19980720>.
+
+
+time:inSeconds a rdf:Property, owl:FunctionalProperty, owl:ReverseFunctionalProperty;
+    rdfs:label "in era seconds";
+    rdfs:domain xsd:dateTime;
```
gkellogg, line 28: Changed to use `xsd:dateTime` instead of `time:ISOTime`.

ns/log.n3
william-vw: The range is `log:Formula`, but it's unclear to me what the relation is between `log:N3Document` and `log:Formula` ..

ns/log.n3
william-vw: Personally I'd leave a caveat here, as to not force N3 systems to support all RDF syntaxes. Something along the lines of "if the RDF syntax is supported, .. , else an error should be raised (?)"

ns/math.n3
william-vw: Unsure whether we need it really ..

ns/math.n3
william-vw: it was unclear what this term meant, until I saw how it was being used :-) yes it seems that this pertains to functional properties ..

ns/math.n3
william-vw: (idem)

ns/time.n3
```diff
@@ -0,0 +1,173 @@
...
+@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .
+@prefix vs: <http://www.w3.org/2003/06/sw-vocab-status/ns#> .
+
+@prefix time: <http://www.w3.org/2000/10/swap/time#>.
+@prefix : <#>.
+
+<> <http://www.w3.org/2000/10/swap/pim/doc.n3#licence> <http://www.w3.org/Consortium/Legal/copyright-software-19980720>.
+
+
+time:inSeconds a rdf:Property, owl:FunctionalProperty, owl:ReverseFunctionalProperty;
```
william-vw, line 26: shouldn't this be owl:InverseFunctionalProperty? ..

ns/crypto.n3
william-vw: Unsure whether they are currently being used ([this CSV](https://github.com/w3c/N3/blob/master/useOfBuiltins/results.csv) seems to imply that they're rarely used)

But, at least in the builtin wishlist there is a request for SHA.

ns/list.n3
william-vw: Certainly the following could be useful (?)
```
( (1 2 3 4) ?l) list:remove (1 2)
```

ns/log.n3
william-vw: Don't think so, personally

ns/log.n3
william-vw: unsure what you mean by this:
>may not bear a good relationship to what is suggested in our builtin wishlist.

gkellogg: These were in the original namespace document, but commented out. I'm not sure if they are the same as suggested in the wish-list document.

ns/log.n3
gkellogg: It's much like the linked-data platform, where there are one or two syntaxes required by the spec, but content-negotiation allows for other formats as well. A processor _should_ process alternate formats, and _must_ evaluate to **false** on failure. In addition to Notation3, formats required to be supported include N-Triples and Turtle?

ns/time.n3
gkellogg:
```suggestion
time:inSeconds a rdf:Property, owl:FunctionalProperty, owl:InverseFunctionalProperty;
```

ns/list.n3
pchampin: This could be achieved by
```
( (1 2) ?l) list:append (1 2 3 4)
```

ns/log.n3
pchampin: The way I read the comments on `log:N3Document` and `log:semantics`, I would interpret `log:N3Document` as something whose `log:content` is valid Notation3. That would make it a subclass of `doc:Work`, which is not explicit above...

william-vw: Do you mean, its contents are an N3 string that needs to be parsed?Yes, that seems to better correspond to its definition: 

> A document which, which parsed as Notation3 as defined in general by http://www.w3.org/DesignIssues/Notation3.html and this schema ...

But, then it's not clear why it wasn't used as domain to `log:content` or `log:semantics`, for instance.

ns/list.n3
william-vw: As you may have inferred from my comment, it didn't actually occur to me that these are new builtins! Sorry - I thought that they were still from the "old" namespace document - since they are still following the previous (and incomplete) way of describing builtins. It's unclear why you're diverging from the more structured builtin description, which we discussed in detail during the meeting.

gkellogg: I was re-purposing the builtin comments already in the namespace documents. Can you provide a reference to the more detailed descriptions? Do they exist for these builtins? I didn't find the information in [Community wish-list of N3 built-ins](https://docs.google.com/document/d/1ByEWSADIvebHRrBda0HhdKhHhp444mYM7BAapXVIJoM/edit?pli=1) too informative, but perhaps there's a more definitive document I'm missing. IMO, we should either get these into the repo, or provide a definitive list of working documents with references in the README.

ns/log.n3
gkellogg: The domain of `log:content` and `log:semantics` is an external document, presumably of type `doc:Work`, although that seems like an archaic term. In the case of `log:content`, the mime-type is expected to be `text/n3` and be valid Notation3 markup. The range of `log:content` is then a `log:N3Document`, as is the domain of `log:parsedAsN3`. The range of both `log:semantics` and `log:parsedAsN3` is a `log:Formula`.

I was confused by the existence of both `log:N3Document` and `log:N3`. Perhaps `log:N3Document` is meant to be a subclass of `doc:Work`? Perhaps the domain of `log:parsedAsN3` and range of `log:n3String` is meant to be `log:N3`.

If `doc:Work` is intended to represent an external document in general, and `log:N3Document` is meant to be an external document with mime-type `text/n3`, then perhaps we need an intermediate class of documents which can be parsed to result in a `log:Formula`, such as JSON-LD and Turtle.

ns/list.n3
gkellogg: Maybe we change the description to something like the following, to allow a more extended use case:

> Iff the subject is a list of two lists, the second component is a subset of the first component and the object is a list composed of the members of the first component of the subject with all members of the second component of the subject removed, matching left to right.
> eg `( (1 2 3 4) (2 3) ) list:remove (1 4)`.
> The object can be calculated as a function of the subject.

ns/log.n3
gkellogg:
```suggestion
log:String  a rdfs:Class; rdfs:subClassOf log:Type, rdfs:Literal.
```

This allows it to be in the range of `log:rawType`, but makes clear that it is a literal.

ns/math.n3
```diff
@@ -0,0 +1,239 @@
...
+@prefix math: <http://www.w3.org/2000/10/swap/math#> .
+@prefix string: <http://www.w3.org/2000/10/swap/string#> .
+@prefix vs: <http://www.w3.org/2003/06/sw-vocab-status/ns#> .
+
+math:StrictProperty a rdfs:Class;
+    rdfs:label "StrictProperty";
+    rdfs:subClassOf rdf:Property.
+
+math:Value a rdfs:Class; rdfs:label "Value"; 
+   rdfs:subClassOf string:String;
```
gkellogg, line 19:
```suggestion
   rdfs:subClassOf [owl:unionOf (xsd:decimal xsd:double xsd:float}]; 
```

This may be too complicated for such a vocabulary, but makes it clear that the type is a numeric type and that a value with this domain can be inferred to be a numeric literal.

ns/math.n3
```diff
@@ -0,0 +1,239 @@
...
+math:StrictProperty a rdfs:Class;
+    rdfs:label "StrictProperty";
+    rdfs:subClassOf rdf:Property.
+
+math:Value a rdfs:Class; rdfs:label "Value"; 
+   rdfs:subClassOf string:String; 
+   rdfs:comment "The class of things that are numeric float values as in Python." .
+
+math:Function a rdfs:Class;
+   rdfs:subClassOf rdf:Property;
```
gkellogg, line 23:
```suggestion
   rdfs:subClassOf rdf:Property, owl:FunctionalProperty;
```

ns/math.n3
```diff
@@ -0,0 +1,239 @@
...
+   rdfs:label "Function";
+   rdfs:comment "A math:Function is unique in terms of math:EqualTo. ".
+
+math:ReverseFunction a rdfs:Class;
+   rdfs:subClassOf rdf:Property;
```
gkellogg, line 28:
```suggestion
   rdfs:subClassOf rdf:Property, owl:InverseFunctionalProperty;
```

ns/math.n3
```diff
@@ -0,0 +1,239 @@
...
+math:LogicalOperator a rdfs:Class; rdfs:subClassOf math:Function, math:ReverseFunction;
+   rdfs:comment """A logical operator allows evaluation eihter way, or testing relationship
+between two values""".
+
+math:List a rdfs:Class; rdfs:label "List"; 
+   rdfs:subClassOf log:List; 
+   rdfs:comment """The class of things that are N3 lists where all of the 
+members are math:Value items.""".
+
+math:TwoMemberedList a rdfs:Class; rdfs:label "TwoMemberedList";
```
gkellogg, line 40: This should probably be in list, and used for the list builtins that have a list with two members.

ns/log.n3
josd: To me yes but in EYE we use --turtle <uri> to get N-Triples or Turtle data:
```
Usage: eye <options>* <data>* <query>*
...
<data>
    [--n3] <uri>                    N3 triples and rules
    --pl <uri>                      Prolog facts and rules
    --proof <uri>                   N3 proof lemmas
    --turtle <uri>                  Turtle triples
...
```

ns/crypto.n3
```diff
@@ -0,0 +1,49 @@
...
+    rdfs:domain rdfs:Literal;
+    rdfs:range xsd:string.
+
+cr:sha a rdf:Property, owl:FunctionalProperty, owl:InverseFunctionalProperty;
+    rdfs:label "SHA-1 hash";
+    rdfs:comment "The object is a SHA-1 hash of the subject.";
+    rdfs:domain rdfs:Literal;
+    rdfs:range xsd:string.
+
+cr:sha256 a rdf:Property, owl:FunctionalProperty, owl:InverseFunctionalProperty;
```
gkellogg, line 39: I took the liberty of defining sha256 and sha512.

ns/list.n3
```diff
@@ -0,0 +1,112 @@
...
+    rdfs:domain log:List;
+    rdfs:comment """
+Iff the subject is a list and the object is in that list, then this is true.
+""";
+    rdfs:isDefinedBy <>.
+
+list:memberAt a rdf:Property;
+    rdfs:label "memberAt";
+    rdfs:domain log:List;
+    rdfs:comment """
```
gkellogg, line 76: Updated from the builtin formalization doc. We could consider using Markdown, and we might split among some different annotation properties.

ns/log.n3
gkellogg: That doesn't scale to other formats (RDF/XML, RDFa, JSON-LD, TriG, ...); we could consider a form with a list subject where additional values are interpreted like an HTTP Accept header to indicate preference. We could also describe a default Accept header (similar to how JSON-LD does) to use to allow the response to be content-negotiated.

We should probably also describe how/(if) an RDF Dataset can be turned into a Formula with recursively quoted graphs.

ns/math.n3
```diff
@@ -0,0 +1,260 @@
...
+- schema
+$a1 math:equalTo $a2
+- summary
+checks equality of numbers
+- definition
+true if and only if $a1 is equal to $a2. 
+requires both arguments to be either concrete numerals, or variables bound to a numeral.
+- literal domains
+$a1: xs:decimal (or its derived types), xs:float, or xs:double  (see note on type promotion, and casting from string)
+$a2: xs:decimal (or its derived types), xs:float, or xs:double  (see note on type promotion, and casting from string)
```
MattesWhite, line 83: Where is this note on type promotion and casting?

ns/list.n3
william-vw: The idea is to have new builtin descriptions that are sufficiently detailed (incl. domains, flow patterns, etc) to allow for consistent implementations - hence our efforts towards [more structured builtin descriptions](https://docs.google.com/document/d/19LSGC7OgGK8oGdPTTrtkgQzYTU5Gq5R-v5FTHf2sHEQ/edit).

ns/log.n3
william-vw: I'd rather say that the range of `log:content` is a xs:string, and its domain is `log:N3Document` (i.e., URIs with that type). Else, `log:N3Document` would be a datatype, no? ..

ns/math.n3
william-vw: Yes, indeed - it also offers a space for tying in our own type promotion and substitution mechanism

ns/log.n3
william-vw: @gkellogg 
> We should probably also describe how/(if) an RDF Dataset can be turned into a Formula with recursively quoted graphs.

I'd leave that for later - see the [note in the spec](https://w3c.github.io/N3/spec/#rel-n3-trig)

ns/math.n3
william-vw: This is very much still a work in progress:
https://docs.google.com/document/d/19LSGC7OgGK8oGdPTTrtkgQzYTU5Gq5R-v5FTHf2sHEQ/edit

ns/log.n3
gkellogg: Note that `log:N3` is defined as a datatype, so it seems to make sense that `log:parsedAsN3` would have that in its domain, and `log:n3String` in it's range.

I updated the definition of `log:N3Document` to make it a subtype of `doc:Work`, but it would overly narrow `log:content` to use that in its domain, as it could presumably retrieve any remote document, not just N3. Same with `log:semantics`, but somewhat in between `doc:Work` and `log:N3Document`, as it could work with Turtle.

gkellogg:
I did some Markdown updates, and use that to generate the HTML versions of the namespace documents.

william-vw:
@gkellogg awesome!!

gkellogg:
I'd suggest we go ahead and merge this so we have a basis for making further changes; it will be easier for people to see when accessed from the site.

Clearly, many more descriptions need to be normalized, and we can add some `skos:example` blocks to many of them, as has been suggested.

Also, the HTML formatting is just a first pass, we can do more to inter-link references, and better normalize formatting, now that it is reduced to a Haml template. The template should probably include some front-matter linking it to the CG.

domel:
I will add SKOS properties e.g. `skos:example`, `skos:note` and will clean the files until Monday.

william-vw:
As per our previous Skype meeting, I added a [builtins.n3 file](https://github.com/w3c/N3/blob/master/files/builtin%20definitions/builtins.n3) that describes allowed inputs for example builtins in (excruciating) detail - in line with the [builtin formalization](https://docs.google.com/document/d/19LSGC7OgGK8oGdPTTrtkgQzYTU5Gq5R-v5FTHf2sHEQ/edit) document (no elaboration on type promotion or substitution). I'm currently testing a simple implementation that loads this file and will validate builtin input accordingly.

gkellogg:
The builtins.n3 file looks promising; do you intend to beef it up to be legitimate OWL, or perhaps N3 rules?

For the case of multiple `:requires`, I think something different is called for to indicate that one of the requirements calls, rather than the implied intersection of requirements.

For example:

```
math:sum
    a :BuiltinDefinition ;
    :subject 
		[ :union ( :numericOrVariableList :variable ) ] ;
    :object :numericOrVariable ;
	:requires [:oneOf ([
		:subject :numericList ;
		:object :numericOrVariable
	] 
	:requires [
		:subject [ :listElementType [ :union ( :maxOneVar :numeric ) ]] ;
		:object :numeric
	])] .
```

william-vw:
Thanks for checking. Yes, you're right re the intersection - it's better to use a union. And, I'm sure there's opportunity for OWL terms. I just want to avoid using classes as type restrictions.

william-vw:
I [updated builtins.n3](https://github.com/w3c/N3/commit/51f370ffb880effb089190076b956d16035f4ea3) to have "unions" for the `requires` relation:

PR #61, opened 2020-11-17
Reasoner tests

gkellogg:
* Remove explicit arc trig functions from tests, relying on reverse operations.
* Remove string version of time tests, and change inSeconds results to be integer, rather than double.
* Add tests for integerQuotient and rounded. Minor vocabulary description update for math:rounded to align with XPath definition.
* Add expected results for conclusion-simple.
* Change concat.n3 to use string:concatenation instead of the obsolete string:concat.
* Add includes/xsd.n3 to test log:outputString. Also adds "string" option to test vocabulary.
* Add includes/conclusion test.
* Use test:conclusions test option instead of test:filter, which means something different. test:conclusions replaces the store with the conclusions of running the rules (similar to log:conclusion).
* Add string:concatenation tests for casting different resource types using XPath casting function.


pchampin (approved):
As you requested feedback during the last call... I made a pass on all commits, and have no objection.

william-vw (commented):
I understand why you're removing them, but I don't recall this being discussed in a CG meeting (maybe I missed it). What if someone's specifically looking for an `asin`, .. builtin and cannot find it? .. (this problem could be resolved with proper docs)

gkellogg:
> I understand why you're removing them, but I don't recall this being discussed in a CG meeting (maybe I missed it). What if someone's specifically looking for an `asin`, .. builtin and cannot find it? .. (this problem could be resolved with proper docs)

IIRC, we decided that the arc-functions were redundant given the bidirectionally of the core functions. But, I can add them back, and we can take this up on a future call.

gkellogg:
> I believe we decided against `integerQuotient` since it's not a "core" builtin - see the builtin wish list:
> 
> https://docs.google.com/document/d/1ByEWSADIvebHRrBda0HhdKhHhp444mYM7BAapXVIJoM/edit#heading=h.gjdgxs

Yes, I missed that.

PR #62, opened 2020-11-17
Update definition of string:concatenation ...

gkellogg:
... to clarify that resources are made into strings using SPARQL/XPath casting function.


ns/string.n3
```diff
@@ -25,15 +25,18 @@ string:concat a owl:InverseFunctionalProperty;   # obj = list of string to strin
     vs:term_status "deprecated";
     owl:inverse string:concatenation.  # Use that instead for new designs
 
-string:concatenation a owl:FunctionalProperty;
+string:concatenation a rdf:Property, owl:FunctionalProperty;
     rdfs:definedBy <>;  
     rdfs:label "is the concatenation of the strings in";
```
pchampin, line 30: This label still suggests that the result is in subject position.
```suggestion
    rdfs:label "concatenate into";
```
I am not entirely sure of this proposal, but a change is required.

gkellogg: As a label, maybe it should be simply "concatenation". If you look at the deprecated `string:concat`, it's label is "is concatenation of", so that would make more sense to me, but "concatenate into" is reasonable, too.

gkellogg:
Note that we previously used the canonical form of literals, but I discovered in some new SPARQL tests that XPath has more specific rules for casting datatypes that result in a more human readable form of literals (particularly floating point). Doing this for N3 keeps us compatible with SPARQL/XPath and makes more sense.

gkellogg:
For tests, see https://github.com/w3c/N3/blob/370427578785bae9ea83ad53d1329bbe01e4b0b5/tests/N3Tests/string/concatenation.n3, which is in PR #61.

william-vw:
We should avoid having two sources of truth here - i.e., the piecemeal updating of builtin N3 files (e.g., `ns/string.n3` and the [builtin google doc](https://docs.google.com/document/d/19LSGC7OgGK8oGdPTTrtkgQzYTU5Gq5R-v5FTHf2sHEQ/edit). I'm unsure how we've evolved on the issues from that document.

gkellogg:
I think the google doc should be either rolled into the wiki, or better into some more normative place. Are the vocabulary documents normative? But, yes, these need to be better synchronized. IMO, the vocabulary documents should become the source of truth, and we could figure out how to include them in a spec, if that's necessary to make them "official".

gkellogg:
I think the builtin google doc is getting pretty unwieldy, given the comments and edits. Also, the vocabulary items reference "casting to/from string" and others, which isn't really appropriate within a namespace document. I'd like to see this either move into the specification, or into a parallel document on builtins. That document can define some of these concepts, and the vocabulary items updated to reference via links. It would be straightforward to use a process similar to turning the n3 vocabulary documents into HTML to create a rollup builtin document with all vocabulary definitions deriving from the n3 source documents, and using regular ReSpec boilerplate. I'd be happy to contribute to making this happen.

Note: if we decide to do this as an appendix to the N3 spec, such a document can be included similar to the way the grammar is included.

william-vw:
> I think the builtin google doc is getting pretty unwieldy, given the comments and edits.

Yes, I can see how this can seem a bit daunting. (But, this is exactly because google docs allow for easier collaborating.)

> I think the google doc should be either rolled into the wiki, or better into some more normative place. 

I think that a normative place would imply that we've somehow evolved on these issues sufficiently (?) Putting them in the wiki is an alternative, it just doesn't allow for commenting or editing with tracked changes.

I think the problem may be that nobody is directly editing the google doc, leaving only strings of comments (unsure why - issues of provenance?). So, I will move the contents to the wiki - I suggest we make this the single source of truth, at least for the time being, until we've agreed on some core issues (such as datatype casting, and input validation).

> It would be straightforward to use a process similar to turning the n3 vocabulary documents into HTML to create a rollup builtin document with all vocabulary definitions deriving from the n3 source documents, and using regular ReSpec boilerplate. I'd be happy to contribute to making this happen.

That sounds great, and I think it'd be extremely useful down the line.

william-vw:
See here for the wiki page on builtin domains:

https://github.com/w3c/N3/wiki/Builtin-domains

gkellogg:
> I think that a normative place would imply that we've somehow evolved on these issues sufficiently (?) Putting them in the wiki is an alternative, it just doesn't allow for commenting or editing with tracked changes.

It's common for editor's drafts to include text which is provisional, as long as it's well marked. You can also tag individual issues for further information. But, it's not easy for non-editors to update the recommendation.

> That sounds great, and I think it'd be extremely useful down the line.

I'll work up an example of such a document in the /ns directory.

PR #63, opened 2020-11-19
Grammar updates for $variables.

gkellogg:
@william-vw I made the changes to the n3.g4 files blind, but they match the EBNF version changes, and work with my parser.

Note that the grammar doesn't say anything about deprecating IRIs as quantified variables.

### DIFF
--- a/grammar/n3.ebnf
+++ b/grammar/n3.ebnf
@@ -45,6 +45,7 @@
 
 [18] pathItem                           ::= iri
                                           | blankNode
+                                          | quantifiedVar
                                           | quickVar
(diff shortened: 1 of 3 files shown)


gkellogg:
Once we agree on this, and merge PR #61, I can update the non-cwm inherited tests to use the '$' syntax; I presume those in N3/cwm_* we'll keep as is, unless there's general support to update these too.

william-vw:
I suppose this has become moot since explicit quantification is no longer supported? 
(On that note, we should update the grammar accordingly ..)

PR #73, opened 2021-03-08
List iterate update

gkellogg:
Updates some general vocabulary generation to be more repeatable, and addresses @william-vw's concern on wording for a ground object value for `list:iterate`.

### DIFF
--- a/ns/index.html
+++ b/ns/index.html
@@ -2,7 +2,7 @@
 <h2>Vocabulary Definitions</h2>
 <section class='vocabulary'>
 <h3>Schema for crypto built-in functions</h3>
-<div class="attribute">
+<div class="attribute comment">
 <p>These are terms drawn to match the available functions
 in <a href="http://freshmeat.sourceforge.net/projects/mxcrypto">mxCrypto</a>.</p>
 
@@ -27,7 +27,7 @@ rdfs:Literal
 </dd>
 <dt>rdfs:range</dt>
 <dd class="range">
-cr:string
+xsd:string
 </dd>
 </dl>
 </dd>
@@ -46,7 +46,7 @@ rdfs:Literal
 </dd>
 <dt>rdfs:range</dt>
 <dd class="range">
-cr:string
+xsd:string
 </dd>
 </dl>
 </dd>
@@ -65,7 +65,7 @@ rdfs:Literal
 </dd>
 <dt>rdfs:range</dt>
 <dd class="range">
-cr:string
+xsd:string
 </dd>
 </dl>
 </dd>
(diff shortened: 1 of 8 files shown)


ns/index.html
```diff
@@ -203,7 +203,9 @@ <h5>Property Definitions</h5>
 If the object is a variable, it will create a solution for each member of the subject list.
 If any member of the object list is a variable, it will create a solution for all matching
 members of the subject list.
-If the object is ground, it evaluates to <code>true</code>; otherwise, <code>false</code>.</p>
+If the object is ground, it evaluates to <code>true</code>
+if the entry at the specified index matches the specified value;
+otherwise, <code>false</code>.</p>
```
TallTed, line 208:
```suggestion
If the object is ground and the entry at the specified index matches the specified value, 
it evaluates to <code>true</code>; otherwise, <code>false</code>.</p>
```

ns/list.html
```diff
@@ -148,7 +148,9 @@ <h5>Property Definitions</h5>
 If the object is a variable, it will create a solution for each member of the subject list.
 If any member of the object list is a variable, it will create a solution for all matching
 members of the subject list.
-If the object is ground, it evaluates to <code>true</code>; otherwise, <code>false</code>.</p>
+If the object is ground, it evaluates to <code>true</code>
+if the entry at the specified index matches the specified value;
+otherwise, <code>false</code>.</p>
```
TallTed, line 153:
```suggestion
If the object is ground and the entry at the specified index matches the specified value, 
it evaluates to <code>true</code>; otherwise, <code>false</code>.</p>
```

ns/list.n3
```diff
@@ -93,7 +93,9 @@ Gets the matching pair of list index and list value for every member of the subj
 If the object is a variable, it will create a solution for each member of the subject list.
 If any member of the object list is a variable, it will create a solution for all matching
 members of the subject list.
-If the object is ground, it evaluates to `true`; otherwise, `false`.
+If the object is ground, it evaluates to `true`
+if the entry at the specified index matches the specified value;
+otherwise, `false`.
```
TallTed, line 98:
```suggestion
If the object is ground and the entry at the specified index matches the specified value, 
it evaluates to <code>true</code>; otherwise, <code>false</code>.</p>
```

gkellogg: Generally use markdown within N3 comments.

gkellogg:
```suggestion
If the object is ground and the entry at the specified index matches the specified value, 
it evaluates to `true`; otherwise, `false`.
```

PR #78, opened 2021-04-16
Paths and other updates

gkellogg:
Includes a number of changes outside of the paths section to better use ReSpec and remove some archaic concepts.

***
<a href="https://pr-preview.s3.amazonaws.com/gkellogg/N3/pull/78.html" title="Last updated on Apr 28, 2021, 11:23 PM UTC (4414081)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/N3/78/9291e3d...gkellogg:4414081.html" title="Last updated on Apr 28, 2021, 11:23 PM UTC (4414081)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -107,7 +107,7 @@
       <p>Below, we elaborate on the main characteristics of <abbr title="Notation3">N3</abbr>:</p>
       <ul>
         <li>
-          <strong><abbr title="Notation3">N3</abbr> is a superset of RDF and <abbr title="Terse RDF Triple Language">Turtle</abbr>.</strong> One can write any valid [[[Turtle]]] graph, and it will be valid in <abbr title="Notation3">N3</abbr> as well. Importantly, this means that all of Turtle's syntactic sugar is available in <abbr title="Notation3">N3</abbr> &mdash; including <a href="#polists">predicate and object lists</a>, unlabeled blank nodes, and <a>collections</a>. Moreover, <strong><a>collections</a> are first-class citizens</strong> in <abbr title="Notation3">N3</abbr>, with an associated set of built-ins for accessing and manipulating them.
+          <strong><abbr title="Notation3">N3</abbr> is a superset of RDF and <abbr title="Terse RDF Triple Language">Turtle</abbr>.</strong> One can write any valid [[[Turtle]]] graph, and it will be valid in <abbr title="Notation3">N3</abbr> as well. Importantly, this means that all of Turtle's syntactic sugar is available in <abbr title="Notation3">N3</abbr> &mdash; including <a href="#polists">predicate and object lists</a>, unlabeled blank nodes, and <a>collections</a>. Moreover, <strong><a>collections</a> are first-class citizens</strong> in <abbr title="Notation3">N3</abbr>, with an associated set of <a>built-ins</a> for accessing and manipulating them.
         </li>
         <li>
           <strong><abbr title="Notation3">N3</abbr> adds an If-Then style of decision making in the form of logical implications and variables.</strong> Logical implications allow making If-Then style inferences via modus ponens (where implementations may apply either backward- or forward-chaining). Variables in such rules may be quantified either universally or existentially; in the latter case, they are comparable to blank nodes. 
@@ -116,10 +116,10 @@
           <strong><abbr title="Notation3">N3</abbr> supports quoting and describing graphs of statements (e.g., recording provenance).</strong> In an open web environment, as an unbounded sea of (semi-)connected sources, one should be made aware of, and given the ability to disseminate, the provenance of information, among other things. A quoted graph includes a conjunction of quoted statements. It allows expression of where the particular statements (e.g., message, document) came from, at what time they were stated and by whom, and, in general, any other description of them.
         </li>
         <li>
-          <strong><abbr title="Notation3">N3</abbr> includes a core set of built-ins for accessing remote online information.</strong> In an open web environment, any online source may have information that supports decision making. The <code>log:semantics</code> built-in allows pulling in (parsed) logical expressions from remote online locations; the <code>log:conclusion</code> built-in allows calculating the deductive closure of any logical expression, whether local or remote. Subsequently, down-stream operations may, for instance, check whether other expressions are included, or not included, within these logical expressions.
+          <strong><abbr title="Notation3">N3</abbr> includes a core set of <a>built-ins</a> for accessing remote online information.</strong> In an open web environment, any online source may have information that supports decision making. The <code>log:semantics</code> <a>built-in</a> allows pulling in (parsed) logical expressions from remote online locations; the <code>log:conclusion</code> <a>built-in</a> allows calculating the deductive closure of any logical expression, whether local or remote. Subsequently, down-stream operations may, for instance, check whether other expressions are included, or not included, within these logical expressions.
         </li>
         <li>
-          <strong><abbr title="Notation3">N3</abbr> supports a scoped version of negation-as-failure.</strong> In an open web environment, it is often useful to check whether or not online information supports, or allows derivation of, a given set of facts. However, it will not be possible nor useful to check whether the whole Semantic Web <em>does not</em> support a given set of facts at some time &mdash; any online source, unknown to the reasoner at the time, or added after the question was asked, may hold a positive answer. Instead, a useful question in this setting is whether a <em>specific</em> piece of information, at a given point in time, does or does not support, or does or does not allow the derivation of, a set of facts. When tightly scoped to a specific information source, and at a specific time, this kind of negation-as-failure will not be influenced by other, unknown online sources. This is referred to as <em>scoped negation as failure,</em> and is supported by <abbr title="Notation3">N3</abbr>'s <code>log:notIncludes</code> built-in.
+          <strong><abbr title="Notation3">N3</abbr> supports a scoped version of negation-as-failure.</strong> In an open web environment, it is often useful to check whether or not online information supports, or allows derivation of, a given set of facts. However, it will not be possible nor useful to check whether the whole Semantic Web <em>does not</em> support a given set of facts at some time &mdash; any online source, unknown to the reasoner at the time, or added after the question was asked, may hold a positive answer. Instead, a useful question in this setting is whether a <em>specific</em> piece of information, at a given point in time, does or does not support, or does or does not allow the derivation of, a set of facts. When tightly scoped to a specific information source, and at a specific time, this kind of negation-as-failure will not be influenced by other, unknown online sources. This is referred to as <em>scoped negation as failure,</em> and is supported by <abbr title="Notation3">N3</abbr>'s <code>log:notIncludes</code> <a>built-in</a>.
         </li>
       </ul>
       <section id='namespaces'>
@@ -607,7 +607,7 @@
           with a <a href="#grammar-production-pathItem">pathItem</a>
           representing a specific <a>N3 triple element</a>
           followed by a path of one or more predicates,
-          separated by directional indicators for
+          separated by <a>directional indicator</a>s for
           forward ("`!`") and reverse ("`^`") relationships.
           In the forward direction, a path describes a relationship from a specific
           <a>N3 triple element</a> (such as an <a>IRI</a>) to a <a>blank node</a>.
@@ -694,7 +694,7 @@
 
       <section id='variables'>
         <h3>Variables and Quantification</h3>
-        <p>Bla...</p>
+        <p class="ednote">Bla...</p>
         <pre class="example" data-transform="updateExample">
           <!--
           Preformatted Text 
@@ -707,7 +707,7 @@
       </section>
       <section id='logimpl'>
         <h3>Logical Implications</h3>
-        <p>Bla...</p>
+        <p class="ednote">Bla...</p>
         <pre class="example nohighlight" data-transform="updateExample">
           <!--
           @prefix foaf: <http://xmlns.com/foaf/0.1/> .
@@ -724,7 +724,7 @@
       </section>
       <section id='keywords'>
         <h4>Keywords and special symbols</h4>
-        <p>Bla...</p>
+        <p class="ednote">Bla...</p>
         <pre class="example nohighlight" data-transform="updateExample">
           <!--
           @prefix foaf: <http://xmlns.com/foaf/0.1/> .
@@ -764,7 +764,7 @@
           <a>lists</a>,
           <a>quoted graphs</a>,
           and of one implication sign.
-          The latter is denoted by the `log:implies` builtin to indicate implication. 
+          The latter is denoted by the `log:implies` <a>built-in</a> to indicate implication. 
           In <abbr title="Notation3">N3</abbr>-turtle syntax, `log:implies` MAY be represented using the symbol <code>=&gt;</code>.
         </p>
         <p>
@@ -836,15 +836,34 @@ To illustrate that consider the following example:
       </section>
       <section id='sinterpret'>
         <h3>Simple N3 interpretations</h3>
-        <p>Bla...</p>
+        <p class="ednote">Bla...</p>
       </section>
       <section id='sneg'>
         <h3>Scoped Negation</h3>
-        <p>Bla...</p>
+        <p class="ednote">Bla...</p>
       </section>
       <section id='builtins'>
-        <h3>N3 Builtins</h3>
-        <p>Bla...</p>
+        <h3>N3 Built-ins</h3>
+        <p><abbr title="Notation3">N3</abbr> defines a core set of <dfn>built-ins</dfn> defined
+          in a set of vocabularies with defined semantics
+          for querying and manipulating <a>N3 documents</a>.
+          Built-ins are denoted by a controlled IRI defined in
+          one of the core namespaces:</p>
+        <ul>
+          <li><a href="../ns/crypto.html">Crypto</a> –
+            <a href="http://www.w3.org/2000/10/swap/crypto#">http://www.w3.org/2000/10/swap/crypto#</a>,</li>
+          <li><a href="../ns/list.html">List</a> –
+            <a href="http://www.w3.org/2000/10/swap/list#">http://www.w3.org/2000/10/swap/list#</a>,</li>
+          <li><a href="../ns/log.html">Log</a> –
+            <a href="http://www.w3.org/2000/10/swap/log#">http://www.w3.org/2000/10/swap/log#</a>,</li>
+          <li><a href="../ns/math.html">Math</a> –
+            <a href="http://www.w3.org/2000/10/swap/math#">http://www.w3.org/2000/10/swap/math#</a>,</li>
+          <li><a href="../ns/string.html">String</a> –
+            <a href="http://www.w3.org/2000/10/swap/string#">http://www.w3.org/2000/10/swap/string#</a>, and</li>
+          <li><a href="../ns/time.html">Time</a> –
+            <a href="http://www.w3.org/2000/10/swap/time#">http://www.w3.org/2000/10/swap/time#</a>.</li>
+        </ul>
+        <p class="ednote">Bla...</p>
       </section>
     </section>
     <section id='n3grammar'>
@@ -970,11 +989,39 @@ To illustrate that consider the following example:
       </section>
       <section id='pathres'>
         <h3>Path resolution</h3>
-        <p>Bla...</p>
+        <p><a>Paths</a> are resolved into zero or more <a>N3 triples</a>
+          and a single <a>N3 triple element</a> which is used as the
+          <a href="#grammar-production-expression">expression</a> value
+          of the <a>path</a>.
+          This is done by recursively processing the <a>path</a>,
+          in reverse, from the
+          last <dfn>directional indicator</dfn> ("`!`") or ("`^`").
+          The result of resolving a <a>path</a> into an
+          <a href="#grammar-production-expression">expression</a>
+          and set of emitted <a>N3 triples</a>
+          MUST be equivalent to using the following algorithm:</p>
+        <ol class="algorithm">
+          <li>If <a>path</a> |p| matches the <a href="#grammar-production-pathItem">pathItem</a> production,
+            the |p| can be reduced no further, return |p|
+            as the result.</li>
+          <li>Otherwise, separate the |p| into two components |path| and |pred|
+            from the last occurance of the <a>directional indicator</a> |dir|.</li>
+          <li>Create |obj| by invoking this algorithm recursively
+            using |path| for |p|.</li>
+          <li>If |dir| is "`!`",
+            emit a new <a>N3 triple</a>
+            (|obj| |pred| |B|) where |B| is a novel <a>blank node</a>.
+            return |obj| as the result.</li>
+          <li>Otherwise, |dir| is "`^`",
+            emit a new <a>N3 triple</a>
+            (|B| |pred| |obj|) where |bn| is a novel <a>blank node</a>.
+            return |B| as the result.</li>
+          </li>
+        </ol>
       </section>
       <section id='grammar'>
         <h3>Grammar</h3>
-        <p>Blah...</p>
+        <p class="ednote">Blah...</p>
         <div data-include="../grammar/n3.html"></div>
         <p>A textual version of this grammar may be found <a href="../grammar/n3.ebnf">here</a>.</p>
       </section>
@@ -1053,7 +1100,7 @@ To illustrate that consider the following example:
           -->
         </pre>
         <p><abbr title="Notation3">N3</abbr> is not directly compatible with TriG as it does not support this graph statement notation. Nevertheless, since <abbr title="Notation3">N3</abbr> supports quoted graphs (i.e., cited formulas) as part of regular <a>N3 statements</a>, authors can utilize the N3 Named Graphs extension [<strong>X</strong>] for associating names or identifiers with cited formulas. Although applications could easily introduce their own custom predicates for this purpose, we strongly recommend the use of this extension for interoperability purposes.</p>
-        <p>The N3 Named Graphs extension [<strong>X</strong>] defines a set of built-ins (used as predicates) to associate names or identifiers with cited formulas, which then become "named graphs". Moreover, each predicate has a well-defined semantics on how the named graph should be interpreted: as quoted graphs (the default <abbr title="Notation3">N3</abbr> interpretation), a partitioning of triples within a dataset context, sets of triples with their own isolated contexts, or specifying relations between local and online graphs.</p>
+        <p>The N3 Named Graphs extension [<strong>X</strong>] defines a set of <a>built-ins</a> (used as predicates) to associate names or identifiers with cited formulas, which then become "named graphs". Moreover, each predicate has a well-defined semantics on how the named graph should be interpreted: as quoted graphs (the default <abbr title="Notation3">N3</abbr> interpretation), a partitioning of triples within a dataset context, sets of triples with their own isolated contexts, or specifying relations between local and online graphs.</p>
       </section>
     </section>
     <section id='patterns' class=informative>
--- a/spec/index.html
+++ b/spec/index.html
@@ -992,50 +992,251 @@ To illustrate that consider the following example:
         <p><a>Paths</a> are resolved into zero or more <a>N3 triples</a>
           and a single <a>N3 triple element</a> which is used as the
           <a href="#grammar-production-expression">expression</a> value
-          of the <a>path</a>.
-          This is done by recursively processing the <a>path</a>,
-          in reverse, from the
-          last <dfn>directional indicator</dfn> ("`!`") or ("`^`").
-          The result of resolving a <a>path</a> into an
-          <a href="#grammar-production-expression">expression</a>
-          and set of emitted <a>N3 triples</a>
-          MUST be equivalent to using the following algorithm:</p>
-        <ol class="algorithm">
-          <li>If <a href="#grammar-production-path">path</a> |p| matches the 
-            <a href="#grammar-production-pathItem">pathItem</a> production, 
-            and no prior reduction took place, 
-            then return |p| as the result.</li>
-          <li>If <a href="#grammar-production-path">path</a> |p| matches the 
-            <a href="#grammar-production-pathItem">pathItem</a> production, 
-            and some prior reduction took place, 
-            then return the last created blank node |B| as the result.</li>
-          <li>Otherwise, separate |p| into two components |pathItem| and |path|, 
-            based on the next occurrence of the <a>directional indicator</a> |dir|.</li>
-          <li>If this is the first reduction of path |p|, 
-          |dir| is "`!`", 
-          and |pathItem2| stands for the next item within the path,
-          emit a new <a>N3 triple</a>
-          (|pathItem| |pathItem2| |B|) where |B| is a novel <a>blank node</a>.
-          </li>
-          <li>If this is the first reduction of path |p|, 
-          |dir| is "`^`", 
-          and |pathItem2| stands for the next item within the path,
-          emit a new <a>N3 triple</a>
-          (|B| |pathItem2| |pathItem|) where |B| is a novel <a>blank node</a>.
-          </li>
-          <li>If this is a subsequent reduction of path |p|, 
-          |dir| is "`!`", 
-          and |B_(n-1)| stands for the previously created blank node,
-          emit a new <a>N3 triple</a>
-          (|B_(n-1)| |pathItem| |B|) where |B| is a novel <a>blank node</a>.
-          </li>
-          <li>If this is a subsequent reduction of path |p|, 
-          |dir| is "`^`", 
-          and |B_(n-1)| stands for the previously created blank node,
-          emit a new <a>N3 triple</a>
-          (|B| |pathItem| |B_(n-1)| where |B| is a novel <a>blank node</a>.
-          </li>
-        </ol>
+          of the <a>path</a>.</p>
+        <p>This section describes two logically equivalent algorithms
+          for transforming a <a>path</a> into a set of <a>N3 triples</a>,
+           and providing a resource to use as the
+          effective
+          <a data-cite="RDF11-CONCEPTS#dfn-subject">subject</a>,
+          <a data-cite="RDF11-CONCEPTS#dfn-predicate">predicate</a>, or
+          <a data-cite="RDF11-CONCEPTS#dfn-object">object</a>
+          in place of the original <a>path</a> expression.</p>
+
+          <section class="algorithm" id="pathres-reverse">
+            <h4>Right to Left Algorithm</h4>
+            <p>The first algorithm describes a means
+              of processing a <a href="#grammar-production-path">path</a>
+              starting with from the right hand side of the <a>path</a>.
+              This is essentially the algorithm described
+              in <a data-cite="N3#path">the Team Submission</a>.</p>
+
+            <p>Processing is performed by
+              recursively processing the <a>path</a> |p|,
+              in reverse, from the
+              last <dfn>directional indicator</dfn> ("`!`") or ("`^`").
+              The result of resolving |p| into an
+              <a href="#grammar-production-expression">expression</a>
+              and set of emitted <a>N3 triples</a>
+              MUST be equivalent to using the following algorithm:</p>
+
+            <ol>
+              <li>If |p| matches the
+                <a href="#grammar-production-pathItem">pathItem</a> production,
+                then |p| can be reduced no further, return |p|
+                as the result.</li>
+              <li>Otherwise, separate the |p| into two components
+                <var>p<sub>n-1</sub></var> and <var>pred<sub>n</sub></var>
+                from the last occurance of the
+                <a>directional indicator</a> <var>dir<sub>n</sub></var>.</li>
+              <li>Create <var>obj<sub>n</sub></var>
+                by invoking this algorithm recursively
+                using <var>p<sub>n-1</sub></var> for |p|.</li>
+              <li>Create a novel <a>blank node</a>
+                <var>B<sub>n</sub></var>.</li>
+              <li>If <var>dir<sub>n</sub></var> is "`!`",
+                emit a new <a>N3 triple</a>
+                (<var>obj<sub>n</sub></var>
+                 <var>pred<sub>n</sub></var>
+                 <var>B<sub>n</sub></var>).</li>
+              <li>Otherwise, <var>dir<sub>n</sub></var> is "`^`",
+                emit a new <a>N3 triple</a>
+                (<var>B<sub>n</sub></var>
+                 <var>pred<sub>n</sub></var>
+                 <var>obj<sub>n</sub></var>).</li>
+              <li>Return <var>B<sub>n</sub></var> as the result.</li>
+            </ol>
+
+            <aside class="example"
+                   title="Processing :jon!:father right to left">
+              <p>Starting with the <a>path</a> |p| set to `:jon!:father`:</p>
+              <ul>
+                <li>Step 1 does not apply, as `:john!:father` is not a
+                  <a href="#grammar-production-pathItem">pathItem</a>.</li>
+                <li>Step 2 separates `:john!:father` into
+                  <var>p<sub>n-1</sub></var> `:john`,
+                  <var>pred<sub>n</sub></var> `:father`, and
+                  <var>dir<sub>n</sub></var> "`!`".</li>
+                <li>Step 3 creates <var>B<sub>n</sub></var>.</li>
+                <li>Step 4 sets <var>obj<sub>n</sub></var>
+                  by calling the algorithm
+                  recursively using |p| `:john`.
+                  <ul>
+                    <li>Step 1 returns `:john` as it matches a
+                      <a href="#grammar-production-pathItem">pathItem</a>.</li>
+                    <li><var>obj<sub>n</sub></var> is `:john`.</li>
+                  </ul>
+                </li>
+                <li>Step 4 matches because <var>dir<sub>n</sub></var> is  "`!`".
+                  The triple (`:john` `:father` <var>B<sub>n</sub></var>) is emitted.</li>
+                <li>Step 6 returns <var>B<sub>n</sub></var>.</li>
+              </ul>
+            </aside>
+
+            <aside class="example"
+                   title="Processing :jon!:father!:mother right to left">
+              <p>Starting with the <a>path</a> |p| set to `jon!:father!:mother`:</p>
+              <ul>
+                <li>Step 1 does not apply, as `jon!:father!:mother` is not a
+                  <a href="#grammar-production-pathItem">pathItem</a>.</li>
+                <li>Step 2 separates `jon!:father!:mother` into
+                  <var>p<sub>n-1</sub></var> `:john!:father`,
+                  <var>pred<sub>n</sub></var> `:mother`, and
+                  <var>dir<sub>n</sub></var> "`!`".</li>
+                <li>Step 3 creates <var>B<sub>n</sub></var>.</li>
+                <li>Step 4 sets <var>obj<sub>n</sub></var> by calling the algorithm
+                  recursively using |p| `:john!:father`.
+                  <ul>
+                    <li>Step 1 does not apply, as `:john!:father` is not a
+                      <a href="#grammar-production-pathItem">pathItem</a>.</li>
+                    <li>Step 2 separates `:john!:father` into
+                      <var>p<sub>n-2</sub></var> `:john`,
+                      <var>pred<sub>n-1</sub></var> `:father`, and
+                      <var>dir<sub>n-1</sub></var> "`!`".</li>
+                    <li>Step 3 creates <var>B<sub>n-1</sub></var>.</li>
+                    <li>Step 4 sets <var>obj<sub>n-1</sub></var>
+                      by calling the algorithm
+                      recursively using |p| `:john`.
+                      <ul>
+                        <li>Step 1 returns `:john` as it matches a
+                          <a href="#grammar-production-pathItem">pathItem</a>.</li>
+                        <li><var>obj<sub>n-1</sub></var> is `:john`.</li>
+                      </ul>
+                    </li>
+                    <li>Step 4 matches because |dir| is  "`!`".
+                      The triple (`:john` `:father` <var>B<sub>n-1</sub></var>) is emitted.</li>
+                    <li>Step 6 returns <var>B<sub>n-1</sub></var>.</li>
+                    <li><var>obj<sub>n</sub></var> is
+                      <var>B<sub>n-1</sub></var>.</li>
+                  </ul>
+                </li>
+                <li>Step 4 matches because |dir| is  "`!`".
+                  The triple
+                  (<var>B<sub>n-1</sub> `:mother` <var>B<sub>n</sub></var>)
+                  is emitted.</li>
+                <li>Step 6 returns <var>B<sub>n</sub></var>.</li>
+              </ul>
+            </aside>
+          </section>
+
+          <section class="algorithm" id="pathres-forward">
+            <h4>Left to Right Algorithm</h4>
+            <p>The second algorithm describes a means
+              of processing a <a href="#grammar-production-path">path</a>
+              starting with from the left hand side of a <a>path</a>.
+              This is useful for an implementation based
+              on the parser productions described in the grammar
+              which can create events in this order.</p>
+
+            <p>Processing is performed by
+              iteratively processing the <a>path</a> <var>p</var>,
+              from the first
+              <a href="#grammar-production-pathItem">pathItem</a>.
+              The result of resolving <var>p</var> into an
+              <a href="#grammar-production-expression">expression</a>
+              and set of emitted <a>N3 triples</a>
+              MUST be equivalent to using the following algorithm:</p>
+
+            <p>Repeat the following algorithm steps starting by
+              initializing |n| to `0`,
+              separating |p| into two components 
+              <var>B<sub>0</sub></var> and
+              <var>p<sub>0</sub></var>
+              based on the next occurrence of the
+              <a>directional indicator</a> <var>dir<sub>0</sub></var>
+              until return.</p>
+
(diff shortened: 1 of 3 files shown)


spec/index.html
```diff
@@ -157,7 +161,7 @@ <h4>Simple triples</h4>
       </section>
       <section id='polists'>
         <h4>Predicate and object lists</h4>
-        <p>As shown in the above example, the same subject (e.g., Spiderman) will often be described by several N3 statements. To make these N3 statements less cumbersome, one can put a '<code>;</code>' at the end of an N3 statement to describe the same subject in the subsequent N3 statement (called a "predicate list"):</p>
+        <p>As shown in the above example, the same subject (e.g., Spiderman) will often be described by several N3 statements. To make these N3 statements less cumbersome, one can put a '<code>;</code>' at the end of an N3 statement to describe the same subject in the subsequent N3 statement (called a "<a href="#polists">predicate list</a>"):</p>
```
TallTed, line 164:
```suggestion
        <p>As shown in the above example, the same subject (e.g., Spiderman) will often be described by several N3 statements. To make these N3 statements less cumbersome, one can put a semicolon ('<code>;</code>') at the end of an N3 statement to describe the same subject in the subsequent N3 statement (called a "<a href="#polists">predicate list</a>"):</p>
```

spec/index.html
```diff
@@ -495,50 +524,58 @@ <h3>Terms</h3>
...
-          <dfn id="quotes">Quoted graphs</dfn> are N3 formulas which are surrounded by quoting signs. 
+          <dfn data-lt="quotes">Quoted graphs</dfn> are N3 formulas which are surrounded by quoting signs. 
           In N3 turtle syntax we denote these quoting signs as curly brackets <code>{}</code>.
         </p>
       </section>
       <section id='formulas'>
         <h3>Formulas</h3>
         <p>
-          An <dfn id="N3-triple">N3 triple</dfn> consists of the three components subject, predicate and object, each being an N3 triple element.
+          An <dfn>N3 triple</dfn> consists of the three components subject, predicate and object, each being an N3 triple element.
```
TallTed, line 566:
```suggestion
          An <dfn>N3 triple</dfn> consists of three components &mdash; subject, predicate, and object &mdash; each being an N3 triple element.
```

spec/index.html
```diff
@@ -495,50 +524,58 @@ <h3>Terms</h3>
...
-          consists of a <a href="#dfn-quantifiers">quantifier</a> <code>Q</code> and an <a data-cite="RDF11-CONCEPTS#section-IRIs">IRI</a> <code>x</code>. 
+          <a>N3 formula</a>
+          f, where a <dfn>quantification</dfn>
+          consists of a <a>quantifier</a> <code>Q</code> and an <a data-cite="RDF11-CONCEPTS#dfn-IRI">IRI</a> <code>x</code>. 
           For a quantified N3 formula <code>Qx f</code>, we call <code>x</code> the quantified element and say that <code>x</code> is quantified in <code>f</code>.
         </p>
         <p>
-          An <dfn id="N3-formula">N3 formula</dfn> is either an <a href="#N3-triple">N3 triple</a>, a <a href="#quantified-formula">quantified N3 formula</a> 
-          or a sequence of <a href="#N3-formula">N3 formulas</a>. 
+          An <dfn>N3 formula</dfn> is either an <a>N3 triple</a>, a <a>quantified N3 formula</a>
```
TallTed, line 577:
```suggestion
          An <dfn>N3 formula</dfn> is either an <a>N3 triple</a>, a <a>quantified N3 formula</a>, 
```

spec/index.html
```diff
@@ -550,7 +587,7 @@ <h2>Semantics</h2>
            <section id='cgraphs'>
         <h3>Closed graphs and ground formulas</h3>
         <p>
-        N3 supports implicit quantification, that is, one can state quantified variables without explicitely stating the quantifier. To do that, we use <a href="#dfn-universals">universal variables</a>, which are implicitely universally quantified, and <a data-cite="RDF11-CONCEPTS#dfn-blank-node">blank nodes</a>, these are implicitely existentially quantified. 
+        N3 supports implicit quantification, that is, one can state quantified variables without explicitely stating the quantifier. To do that, we use <a>universal variables</a>, which are implicitely universally quantified, and <a data-cite="RDF11-CONCEPTS#dfn-blank-node">blank nodes</a>, these are implicitely existentially quantified.
```
TallTed, line 590:
```suggestion
        N3 supports implicit quantification; that is, one can state quantified variables without explicitly stating the quantifier. To do that, we use <a>universal variables</a>, which are implicitly universally quantified, and <a data-cite="RDF11-CONCEPTS#dfn-blank-node">blank nodes</a>, which are implicitly existentially quantified. 
```

spec/index.html
```diff
@@ -740,7 +776,7 @@ <h3>SPARQL</h3>
         <p>Below, we indicate some important differences between this production and N3:</p>
         <ul>
           <li>
-            Like N3, SPARQL permits literals as the subject of RDF triples. But, in contrast to N3, it does not allow literals as the predicate of RDF triples. Similarly, N3 allows for blank node property lists and collections in any position, whereas SPARQL only allows them in the subject or object position.
+            Like N3, SPARQL permits literals as the subject of RDF triples. But, in contrast to N3, it does not allow literals as the predicate of RDF triples. Similarly, N3 allows for <a href="#bnodeprplist">blank node property lists</a> and collections in any position, whereas SPARQL only allows them in the subject or object position.
```
TallTed, line 779:
```suggestion
            Like N3, SPARQL permits literals as the subject of RDF triples, but, in contrast to N3, it does not allow literals as the predicate of RDF triples. Similarly, N3 allows for <a href="#bnodeprplist">blank node property lists</a> and collections in any position, whereas SPARQL only allows them in the subject or object position.
```

spec/index.html
```diff
@@ -187,7 +191,7 @@ <h3>IRIs</h3>
 
           :green-goblin :enemyOf :spiderman .
         </pre>
-        <p>One can write <dfn id="dfn-relative-iri">relative IRIs</dfn>, e.g., <code>&lt;#Spiderman&gt;</code>, which will be resolved relative to the <em>current base IRI</em>. In other words, a relative IRI is turned into an IRI by concatenating the base IRI with the local name (e.g., <code>Spiderman</code>). A base IRI can be defined using the '@base' directive. For instance, the following is equivalent to the prior example:</p>
+        <p>One can write <dfn data-lt="relative iri">relative IRIs</dfn>, e.g., <code>&lt;#Spiderman&gt;</code>, which will be resolved relative to the <em>current base IRI</em>. In other words, a relative IRI is turned into an IRI by concatenating the base IRI with the local name (e.g., <code>Spiderman</code>). A base IRI can be defined using the '@base' directive. For instance, the following is equivalent to the prior example:</p>
```
TallTed, line 194:
```suggestion
        <p>One can write <dfn data-lt="relative iri">relative IRIs</dfn>, e.g., <code>&lt;#Spiderman&gt;</code>, which will be resolved relative to the <em>current base IRI</em>. In other words, a relative IRI is turned into an IRI by concatenating the base IRI with the local name (e.g., <code>Spiderman</code>). A base IRI can be defined using the `@base` directive. For instance, the following is equivalent to the prior example:</p>
```

spec/index.html
```diff
@@ -295,7 +299,7 @@ <h3>Blank Nodes</h3>
             There likely already exists an IRI that identifies the resource, and we don't want to mint a new IRI for a resource that likely already has one (nor do we want to do an extensive search to find out what that IRI is).
           </li>
         </ul>
-        <p>Instead, you can use <dfn id="dfn-blank-nodes">blank nodes</dfn> to talk about resources. They are existential variables, i.e., stating the existence of a thing without identifying that thing. (In fact, a blank node is equivalent to an explicit <a href="#variables">existential variable</a> in N3, aside from some scoping issues). Blank nodes can be represented in several ways:</p>
+        <p>Instead, you can use <dfn data-cite="RDF11-CONCEPTS#dfn-blank-node">blank nodes</dfn> to talk about resources. They are existential variables, i.e., stating the existence of a thing without identifying that thing. (In fact, a blank node is equivalent to an explicit <a href="#variables">existential variable</a> in N3, aside from some scoping issues). Blank nodes can be represented in several ways:</p>
```
TallTed, line 302:
```suggestion
        <p>Instead, you can use <dfn data-cite="RDF11-CONCEPTS#dfn-blank-node">blank nodes</dfn> to talk about resources. They are existential variables; i.e., they state the existence of a thing without identifying that thing. (In fact, aside from some scoping issues, a blank node is equivalent to an explicit <a href="#variables">existential variable</a> in N3.) Blank nodes can be represented in several ways:</p>
```

spec/index.html
```diff
@@ -495,50 +524,58 @@ <h3>Terms</h3>
...
+          The set of <dfn data-lt="triple element|n3 triple element">N3 triple elements</dfn> consists of the disjoint sets of
+          <a data-cite="RDF11-CONCEPTS#dfn-iri">IRIs</a>, 
+          <a data-cite="RDF11-CONCEPTS#dfn-blank-node">blank nodes</a>,
+          <a data-cite="RDF11-CONCEPTS#dfn-literal">literals</a>,  
+          <a>universal variables</a>,
+          <a>lists</a>,
+          <a>quoted graphs</a>,
+          and of one implication sign.
+          The latter is denoted by the `log:implies` builtin to indicate implication. 
+          In N3-turtle syntax `log:implies` MAY be represented using the symbol <code>=&gt;</code>.
```
TallTed, line 540:
```suggestion
          In N3-turtle syntax, `log:implies` MAY be represented using the symbol <code>=&gt;</code>.
```

TallTed (changes requested):
Doubtless, I've missed some things ... and I only looked at the changed lines here.

gkellogg:
Work continues. This has turned into a larger update than I had originally intended (see commit comments), I hope I'm not overstepping my mandate.

We need to describe _verbs_ more generally, and have a resolution section in the EBNF Grammar section. My thought is to describe things that are beyond Turtle as transformations to an equivalent Turtle structure (where feasible). We'll need more on parsing Graphs, and need to decide if we're changing from the use of "Formulae" to "Formulas"; the Team Submission used "Formulae" as the plural of "Formulas".

spec/index.html
```diff
@@ -80,28 +103,28 @@
...
-      <p>Notation3 Logic, or N3 for short, aims to improve decision-making abilities in an open web environment - by (a) extending the logical representation abilities of RDF and (b) allowing to access, operate on, and reason over online information. In line with the design ethos of RDF, N3 attempts to walk the line between, on the one hand, ease-of-use by authors and simplicity of reasoning for developers; and, on the other hand, an extended utility and practicality for building real-world applications.</p>
-      <p>Below, we elaborate on the main characteristics of N3:</p>
+      <p>Notation3 Logic, or <abbr title="Notation3">N3</abbr> for short, aims to improve decision-making abilities in an open web environment &mdash; by (a) extending the logical representation abilities of RDF and (b) allowing to access, operate on, and reason over online information. In line with the design ethos of RDF, <abbr title="Notation3">N3</abbr> attempts to walk the line between, on the one hand, ease-of-use by authors and simplicity of reasoning for developers; and, on the other hand, an extended utility and practicality for building real-world applications.</p>
```
TallTed, line 106:
```suggestion
      <p>Notation3 Logic, or <abbr title="Notation3">N3</abbr> for short, aims to improve decision-making abilities in an open web environment &mdash; by (a) extending the logical representation abilities of RDF and (b) allowing access to, operation on, and reasoning over online information. In line with the design ethos of RDF, <abbr title="Notation3">N3</abbr> attempts to walk the line between, on the one hand, ease-of-use by authors and simplicity of reasoning for developers; and, on the other hand, extended utility and practicality for building real-world applications.</p>
```

spec/index.html
```diff
@@ -80,28 +103,28 @@
...
       <ul>
         <li>
-          <strong>N3 is a superset of RDF and <abbr title="Terse RDF Triple Language">Turtle</abbr>.</strong> One can write any valid [[[turtle]]] graph and it will be valid in N3 as well. Importantly, this means that all of Turtle's syntactic sugar is available in N3 - including predicate and object lists, unlabeled blank nodes, and collections. Moreover, <strong>collections are first-class citizens</strong> in N3, with an associated set of built-ins for accessing and manipulating them.
+          <strong><abbr title="Notation3">N3</abbr> is a superset of RDF and <abbr title="Terse RDF Triple Language">Turtle</abbr>.</strong> One can write any valid [[[turtle]]] graph and it will be valid in <abbr title="Notation3">N3</abbr> as well. Importantly, this means that all of Turtle's syntactic sugar is available in <abbr title="Notation3">N3</abbr> &mdash; including <a href="#polists">predicate and object lists</a>, unlabeled blank nodes, and <a>collections</a>. Moreover, <strong><a>collections</a> are first-class citizens</strong> in <abbr title="Notation3">N3</abbr>, with an associated set of built-ins for accessing and manipulating them.
```
TallTed, line 110:
```suggestion
          <strong><abbr title="Notation3">N3</abbr> is a superset of RDF and <abbr title="Terse RDF Triple Language">Turtle</abbr>.</strong> One can write any valid [[[Turtle]]] graph, and it will be valid in <abbr title="Notation3">N3</abbr> as well. Importantly, this means that all of Turtle's syntactic sugar is available in <abbr title="Notation3">N3</abbr> &mdash; including <a href="#polists">predicate and object lists</a>, unlabeled blank nodes, and <a>collections</a>. Moreover, <strong><a>collections</a> are first-class citizens</strong> in <abbr title="Notation3">N3</abbr>, with an associated set of built-ins for accessing and manipulating them.
```

spec/index.html
```diff
@@ -80,28 +103,28 @@
...
         </li>
         <li>
-          <strong>N3 adds an If-Then style of decision making in the form of logical implications and variables.</strong> Logical implications allow making If-Then style inferences via modus ponens (where implementations may apply either backward- or forward-chaining). Variables in such rules may be either universally or existentially quantified - in the latter case, they are comparable to blank nodes. 
+          <strong><abbr title="Notation3">N3</abbr> adds an If-Then style of decision making in the form of logical implications and variables.</strong> Logical implications allow making If-Then style inferences via modus ponens (where implementations may apply either backward- or forward-chaining). Variables in such rules may be either universally or existentially quantified &mdash; in the latter case, they are comparable to blank nodes.
```
TallTed, line 113:
```suggestion
          <strong><abbr title="Notation3">N3</abbr> adds an If-Then style of decision making in the form of logical implications and variables.</strong> Logical implications allow making If-Then style inferences via modus ponens (where implementations may apply either backward- or forward-chaining). Variables in such rules may be quantified either universally or existentially; in the latter case, they are comparable to blank nodes. 
```

spec/index.html
```diff
@@ -80,28 +103,28 @@
...
         </li>
         <li>
-          <strong>N3 supports quoting and describing graphs of statements (e.g., recording provenance).</strong> In an open web environment, as an unbounded sea of (semi-)connected sources, one should be made aware of, and given the ability to disseminate, the provenance of information, among others. A quoted graph includes a conjunction of quoted statements. It allows expressing where the particular statements (e.g., message, document) came from, at what time they were stated and by whom, and, in general, what other kinds of statements can be made about them.
+          <strong><abbr title="Notation3">N3</abbr> supports quoting and describing graphs of statements (e.g., recording provenance).</strong> In an open web environment, as an unbounded sea of (semi-)connected sources, one should be made aware of, and given the ability to disseminate, the provenance of information, among others. A quoted graph includes a conjunction of quoted statements. It allows expressing where the particular statements (e.g., message, document) came from, at what time they were stated and by whom, and, in general, what other kinds of statements can be made about them.
```
TallTed, line 116:
```suggestion
          <strong><abbr title="Notation3">N3</abbr> supports quoting and describing graphs of statements (e.g., recording provenance).</strong> In an open web environment, as an unbounded sea of (semi-)connected sources, one should be made aware of, and given the ability to disseminate, the provenance of information, among other things. A quoted graph includes a conjunction of quoted statements. It allows expression of where the particular statements (e.g., message, document) came from, at what time they were stated and by whom, and, in general, any other description of them.
```

spec/index.html
```diff
@@ -80,28 +103,28 @@
...
+          <strong><abbr title="Notation3">N3</abbr> includes a core set of builtins for accessing remote online information.</strong> In an open web environment, any online source may have information to support decision making. The <code>log:semantics</code> builtin allows pulling in (parsed) logical expressions from remote online locations; the <code>log:conclusion</code> built-in allows calculating the deductive closure of any logical expression, albeit local or remote. Subsequently, down-stream operations may, for instance, check whether other expressions are included, or not included, within these logical expressions.
```
TallTed, line 119:
```suggestion
          <strong><abbr title="Notation3">N3</abbr> includes a core set of built-ins for accessing remote online information.</strong> In an open web environment, any online source may have information that supports decision making. The <code>log:semantics</code> built-in allows pulling in (parsed) logical expressions from remote online locations; the <code>log:conclusion</code> built-in allows calculating the deductive closure of any logical expression, whether local or remote. Subsequently, down-stream operations may, for instance, check whether other expressions are included, or not included, within these logical expressions.
```

spec/index.html
```diff
@@ -80,28 +103,28 @@
...
+          <strong><abbr title="Notation3">N3</abbr> supports a scoped version of negation-as-failure.</strong> In an open web environment, it is often useful to check whether online information supports or not, or allows to derive or not, a given set of facts. But, it will not be possible nor useful to check whether the whole Semantic Web <em>does not</em> support a given set of facts at some time &mdash; any online source, unknown to the reasoner at the time, or added after the question was asked, may hold a positive answer. Instead, a useful question in this setting is whether a <em>specific</em> piece of information, at a given point in time, does not support, or does not allow the derivation of, a set of facts. When tightly scoped to a specific information source, and at a specific time, this kind of negation-as-failure will not influenced by other, unknown online sources. This is referred to as scoped negation as failure, and is supported by <abbr title="Notation3">N3</abbr>'s <code>log:notIncludes</code> built-in.
```
TallTed, line 122:
```suggestion
          <strong><abbr title="Notation3">N3</abbr> supports a scoped version of negation-as-failure.</strong> In an open web environment, it is often useful to check whether or not online information supports, or allows derivation of, a given set of facts. However, it will not be possible nor useful to check whether the whole Semantic Web <em>does not</em> support a given set of facts at some time &mdash; any online source, unknown to the reasoner at the time, or added after the question was asked, may hold a positive answer. Instead, a useful question in this setting is whether a <em>specific</em> piece of information, at a given point in time, does or does not support, or does or does not allow the derivation of, a set of facts. When tightly scoped to a specific information source, and at a specific time, this kind of negation-as-failure will not be influenced by other, unknown online sources. This is referred to as <em>scoped negation as failure,</em> and is supported by <abbr title="Notation3">N3</abbr>'s <code>log:notIncludes</code> built-in.
```

spec/index.html
```diff
@@ -132,73 +155,106 @@ <h3>Namespaces</h3>
 
     <section id='language' class=informative>
       <h2>Language</h2>
-      <p>The aim of this section is to provide an informal overview of the N3 language and its different features. Where possible, this section borrows from, or is at least based on, the [[[turtle]]] specification. More formal definitions will follow in the subsequent sections.
+      <p>The aim of this section is to provide an informal overview of the <abbr title="Notation3">N3</abbr> language and its different features. Where possible, this section borrows from, or is at least based on, the [[[turtle]]] specification. More formal definitions will follow in the subsequent sections.
```
TallTed, line 158:
```suggestion
      <p>The aim of this section is to provide an informal overview of the <abbr title="Notation3">N3</abbr> language and its different features. Where possible, this section borrows from, or is at least based on, the [[[Turtle]]] specification. More formal definitions will follow in the subsequent sections.
```

spec/index.html
```diff
@@ -132,73 +155,106 @@ <h3>Namespaces</h3>
...
       </p>
       <section id='n3doc'>
         <h3>N3 document</h3>
-        <p>An N3 document represents an N3 graph in a compact textual form. An N3 graph, in its simplest form, is a series of <a href="#simpletriples">N3 statements or triples</a>, each describing a resource's property and written as a triple: <code>&lt;subject&gt; &lt;predicate&gt; &lt;object&gt;.</code> An N3 graph can have many more elements - <a href="#cformulas">cited formulas</a> for quoting sets of statements, <a href="#logimpl">logical implications</a> for If-Then style, rule-based reasoning, <a href="#variables">quantifications</a> for variables, and so on. We introduce these elements below.</p>
-        <p>Comments are indicated using a separate '#' (i.e., not part of another N3 symbol, such as an IRI) and continue until the end of 
-          the line.
+        <p>An <dfn>N3 document</dfn> represents an <a>N3 graph</a> in a compact textual form. An <dfn>N3 graph</dfn>,
+          in its simplest form, is a series of <dfn data-lt="n3 triple|statement|triple">N3 statements</dfn> or <a>triples</a>,
+          each describing a resource's relationships and written as a triple:
```
TallTed, line 164:
```suggestion
          each describing one relationship of a resource, and written as a triple:
```

spec/index.html
```diff
@@ -132,73 +155,106 @@ <h3>Namespaces</h3>
...
+          <a data-cite="RDF11-CONCEPTS#dfn-subject">subject</a>,
+          <a data-cite="RDF11-CONCEPTS#dfn-predicate">predicate</a>, and
+          <a data-cite="RDF11-CONCEPTS#dfn-object">object</a>.
+          An <a>N3 graph</a> can have many more elements -
```
TallTed, line 168:
```suggestion
          An <a>N3 graph</a> can have many more elements &mdash;
```

spec/index.html
```diff
@@ -132,73 +155,106 @@ <h3>Namespaces</h3>
...
+          <a data-cite="RDF11-CONCEPTS#dfn-subject">subject</a>, 
+          <a data-cite="RDF11-CONCEPTS#dfn-predicate">predicate</a>, and
+          <a data-cite="RDF11-CONCEPTS#dfn-object">object</a>,
+          separated by whitespace and terminated by '`.`'
+          after each <a>triple</a>.
+          For now, we will asssume that a resource is represented either by an <a href="#iris">IRI</a> or a <a href="#literals">literal</a>.
+          In general, one uses an IRI to identify, well, an identifiable entity,
+          such as a person, place, or thing &mdash;
+          a literal is used for a textual or numerical (i.e., datatype) value,
+          such as names, dates, height, and so on.</p>
```
TallTed, line 191:
```suggestion
          such as name, date, height, and so on.</p>
```

spec/index.html
```diff
@@ -132,73 +155,106 @@ <h3>Namespaces</h3>
...
+        <p>The three <a>N3 triples</a> below highlights the emnity between Spiderman and the Green Goblin, and lists the names of these two individuals:</p>
```
TallTed, line 192:
```suggestion
        <p>The three <a>N3 triples</a> below highlight the enmity between Spiderman and the Green Goblin, and lists the names of these two individuals:</p>
```

spec/index.html
```diff
@@ -132,73 +155,106 @@ <h3>Namespaces</h3>
...
+          <!--
+          <http://example.org/#spiderman> <http://example.org/#enemyOf> <http://example.org/#green-goblin> .
+          <http://example.org/#spiderman> <http://xmlns.com/foaf/0.1/name> "Spiderman" .
+          <http://example.org/#green-goblin> <http://xmlns.com/foaf/0.1/name> "Green Goblin" .
+            -->
         </pre>
-        <p class="note">There is no inherent order for N3 / RDF triples. This is the same for the relational model, i.e., relational tuples or rows do not have an inherent order. Hence, it is a mistake to associate meaning with the order of a statement in an N3 document: e.g., assuming that the first listed telephone for Spiderman would his landline, and the second one his mobile number.</p>
+        <p class="note">There is no inherent order for <abbr title="Notation3">N3</abbr> / RDF triples. This is the same for the relational model, i.e., relational tuples or rows do not have an inherent order. Hence, it is a mistake to associate meaning with the order of a statement in an <abbr title="Notation3">N3</abbr> document: e.g., assuming that the first listed telephone for Spiderman would his landline, and the second one his mobile number.</p>
```
TallTed, line 205:
```suggestion
        <p class="note">There is no inherent order for <abbr title="Notation3">N3</abbr> or RDF triples. The same is true for the relational model, i.e., relational tuples or rows do not have an inherent order. Hence, it is a mistake to associate meaning with the order of statements in an <abbr title="Notation3">N3</abbr> document; for instance, it would be an error to assume that the first listed telephone for Spiderman would be his landline, and the second one his mobile number.</p>
```

spec/index.html
```diff
@@ -132,73 +155,106 @@ <h3>Namespaces</h3>
...
       </section>
       <section id='polists'>
         <h4>Predicate and object lists</h4>
-        <p>As shown in the above example, the same subject (e.g., Spiderman) will often be described by several N3 statements. To make these N3 statements less cumbersome, one can put a semicolon ("<code>;</code>") at the end of an N3 statement to describe the same subject in the subsequent N3 statement (called a "<a href="#polists">predicate list</a>"):</p>
-        <pre class="example nohighlight">
-          &lt;http://example.org/#spiderman&gt;
-              &lt;http://example.org/#enemyOf&gt; &lt;http://example.org/#green-goblin&gt; ;
-              &lt;http://xmlns.com/foaf/0.1/name&gt; "Spiderman" .
+        <p>As shown in the above example, the same subject (e.g., Spiderman) will often be described by several <a>N3 statements</a>. To make these <a>N3 statements</a> less cumbersome, one can put a semicolon ("<code>;</code>") at the end of an <a>N3 statement</a> to describe the same subject in the subsequent <a>N3 statement</a> (called a "<a href="#polists">predicate list</a>"):</p>
```
TallTed:
```suggestion
        <p>As shown in the above example, the same subject (here, Spiderman) will often be described by several <a>N3 statements</a>. To make these <a>N3 statements</a> less cumbersome, one can put a semicolon ("<code>;</code>") at the end of an <a>N3 statement</a> to describe the same subject in the subsequent <a>N3 statement</a> (called a "<a href="#polists">predicate list</a>"):</p>
```

spec/index.html
```diff
@@ -132,73 +155,106 @@ <h3>Namespaces</h3>
...
-        <p class="note">N3 also supports case-insensitive 'PREFIX' and 'BASE' directives, as does Turtle, to align the syntax with SPARQL. Note that the <code>@prefix</code> and <code>@base</code> directives require a trailing '<code>.</code>' after the IRI, but the equivalent<code>PREFIX</code> and <code>BASE</code> must not have a trailing '.'.
+        <p class="note">We recommend listing the <code>@prefix</code> declarations (just like <code>@base</code> declarations) at the top of an <a>N3 document</a>. This is not mandatory however, and they can technically be put anywhere before the prefixed name or relative IRI that relies on the declaration. Subsequent <code>@prefix</code> or <code>PREFIX</code> directives may re-map the same prefix label to another namespace IRI.</p>
```
TallTed:
```suggestion
        <p class="note">We recommend listing the <code>@prefix</code> declarations (just like <code>@base</code> declarations) at the top of an <a>N3 document</a>. This is not mandatory, however, and they can technically be put anywhere before the prefixed name or relative IRI that relies on the declaration. Subsequent <code>@prefix</code> or <code>PREFIX</code> directives may re-map the same prefix label to another namespace IRI.</p>
```

spec/index.html
```diff
@@ -209,41 +265,49 @@ <h3>IRIs</h3>
       </section>
       <section id='literals'>
         <h3>Literals</h3>
-        <p>In general, one uses a literal to represent a textual or numerical (i.e., datatype) value - such as names, dates, height, and so on. Numbers (integers, decimals and doubles) can simply be represented using their numerical value, and booleans can be represented using keywords <code>true</code> or <code>false</code>:</p>
-        <pre class="example nohighlight">
-            @prefix : &lt;http://example.org/elements&gt; .
-            @prefix rdfs: &lt;http://www.w3.org/2000/01/rdf-schema#&gt; .
+        <p>In general, one uses a literal to represent a textual or numerical (i.e., datatype) value &mdash; such as names, dates, height, and so on. Numbers (integers, decimals and doubles) can simply be represented using their numerical value, and booleans can be represented using keywords <code>true</code> or <code>false</code>:</p>
```
TallTed, line 302:
```suggestion
        <p>In general, one uses a literal to represent a textual or numerical (i.e., datatype) value &mdash; such as name, date, height, and so on. Numbers (integers, decimals, and doubles) can simply be represented using their numerical value, and booleans can be represented using keywords <code>true</code> or <code>false</code>:</p>
```

gkellogg:
> Doubtless, I've missed some things ... and I only looked at the changed lines here.

Thanks, Ted. Most of my work has been organizational, without trying to do too much stylistically or grammatically so far. But, you're the best on this.

gkellogg:
Work is now complete and can be considered for merging.

Note that the GitHub Action will fail due to a problem introduced in ReSpec, which will presumably be resolved soon (inappropriate use of aria attributes).

william-vw (changes requested):
See commit e06a0b6 (i.e. on how I think the path algorithm should work)

TallTed (changes requested):

spec/index.html
```diff
@@ -1001,21 +1001,32 @@ <h3>Path resolution</h3>
...
-            emit a new <a>N3 triple</a>
-            (|obj| |pred| |B|) where |B| is a novel <a>blank node</a>.
-            return |obj| as the result.</li>
-          <li>Otherwise, |dir| is "`^`",
-            emit a new <a>N3 triple</a>
-            (|B| |pred| |obj|) where |bn| is a novel <a>blank node</a>.
-            return |B| as the result.</li>
+          <li>If <a href="#grammar-production-path">path</a> |p| matches the <a href="#grammar-production-pathItem">pathItem</a> production, and no prior reduction took place, then return |p| as the result.</li>
+          <li>If <a href="#grammar-production-path">path</a> |p| matches the <a href="#grammar-production-pathItem">pathItem</a> production, and some prior reduction took place, then return the last created blank node |B| as the result.</li>
+          <li>Otherwise, separate |p| into two components |pathItem| and |path|, based on the next occurrence of the <a>directional indicator</a> |dir|.</li>
```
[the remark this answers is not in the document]
gkellogg: I agree, and I generally try to break lines on natural phrasing.

gkellogg:
> See commit [e06a0b6](https://github.com/w3c/N3/commit/e06a0b6f9976898c9e700567cd654da51ad97155) (i.e. on how I think the path algorithm should work)

I have to say, I don't find your version of the algorithm clearer; the recursive nature of what I proposed I find easier to understand, but perhaps there's a middle ground.

Stylistically, it's good form to reduce line lengths (for the reasons @TallTed stated) and use "Otherwise" for the else-side of a clause.

If the concern is for correctness, then we should also re-consider the "Paths" section, which informally defines the same transitions as re-writing the input. So that `:joe!:mother` becomes `[is :mother of :joe]` or `:joe :mother B`. I'll examine this further, but it does seem that the result will always be a blank node, and not _obj_ in the case of my step 4.

My own processor does not process paths in reverse, and implements something closer to what you state, but as a description, reverse processing seems to be easier to follow, and the intent is the results, not the method. Describing it as a forward process becomes more cumbersome, even if this is the AST generated by a parser.

TallTed:
@william-vw --
> See commit [e06a0b6](https://github.com/w3c/N3/commit/e06a0b6f9976898c9e700567cd654da51ad97155) (i.e. on how I think the path algorithm should work)

Would you please make a complete PR from your fork, so that the preview of your fork can be compared to the preview of @gkellogg's fork?  Or make a change suggestion on @gkellogg's PR, so that it's (somewhat) more easily seen where the differences in the source are?

william-vw:
@gkellogg It's not an issue of clarity but rather correctness.. (But, as said, I could be wrong since some terms are unclear to me, e.g., |pred|, which does not occur in the grammar)

As I understood it, given the following:
```
:john!:father a :Person .
```
Then I believe the recursive call would return the last `:father`, and the algorithm would use it as the subject of the statement 
```
:father :father _:bn0
```
Instead of
```
:john :father _:bn0
```

Honestly I don't think a recursive process is the best representation here, since the next bnode would never be utilized in the prior emitted statement. E.g., for
```
:john!:father!:father a :Person .
```
Should yield
```
:john :father _:bn0 . _:bn0 :father _:bn1 . _:bn1 a :Person .
```
And not
```
:john :father _:bn1 . _:bn0 :father _:bn1 . _:bn1 a :Person .
```

Note that the first item in the path should also be treated differently: `:john` should act as the subject of the first emitted statement, whereas the other path items should act as predicates.

gkellogg:
> As I understood it, given the following:
> 
> ```
> :john!:father a :Person .
> ```
> 
> Then I believe the recursive call would return the last `:father`, and the algorithm would use it as the subject of the statement
> 
> ```
> :father :father _:bn0
> ```
> 
> Instead of
> 
> ```
> :john :father _:bn0
> ```

Following my algorithm, you would get the following:

* Step 1 does not apply, as `:john!:father` is not a pathItem.
* Step 2 separate `:john!:father` into _path_=`:john`, _pred_=`:father`, and _dir_=`!`.
* Step 3 invokes the algorithm recursively using `:john`:
  * Step 1 returns `:john` as it matches a pathItem.
  * _obj_=`:john`.
* Step 4 matches because _dir_=`!`. The triple `(:john :father B)` is emitted, and `:john` is returned as a result.
  > As I said before, I believe this is wrong, and `B` should be returned.

As the path was replaced with `B`, the remaining statement is `(B a :Person)`. In fact, my original algorithm would have resulted in `(:john a :Person)`, which is incorrect.



> Honestly I don't think a recursive process is the best representation here, since the next bnode would never be utilized in the prior emitted statement. E.g., for
> 
> ```
> :john!:father!:father a :Person .
> ```

This is not the a characteristic of the recursive algorithm, but of the misappropriation of _obj_ as the result, and not `B`. Following the corrected flow of my algorithm you'd get the following steps:

* Step 1 does not apply, as `:john!:father!:father` is not a pathItem.
* Step 2 separate `:john!:father!:father` into _path_=`:john!:father`, _pred_=`:father`, and _dir_=`!`.
* Step 3 invokes the algorithm recursively using `:john!:father`:
  * Step 1 does not apply, as `:john!:father` is not a pathItem.
  * Step 2 separate `:john!:father` into _path_=`:john`, _pred_=`:father`, and _dir_=`!`.
  * Step 3 invokes the algorithm recursively using `:john`:
    * Step 1 returns `:john` as it matches a pathItem.
  * _obj_=`:john`.
  * Step 4 matches because _dir_=`!`. The triple `(:john :father B0)` is emitted, and `B0` is returned as a result.
* _obj_=`B0`
* Step 4 matches because _dir_=`!`. The triple `(B0 :father B1)` is emitted, and `B1` is returned as a result.

As the path was replaced with `B1`, the remaining statement is `(B1 a :Person)`.

So, the resulting triples are: `:john :father [:father [a :Person]] .`, which I believe is correct.

@TallTed said:
> Would you please make a complete PR from your fork, so that the preview of your fork can be compared to the preview of @gkellogg's fork?

This would be useful: you can actually make a PR to the branch that this PR is based. Doing it through the suggestion mechanism is probably possible, but challenging. I'll just lay them out here in Markdown.

## Gregg's (corrected) algorithm
1. If `path` |p| matches the `pathItem` production, then |p| can be reduced no further, return |p| as the result.
2. Otherwise, separate |p| into two components |path| and |pred| from the last occurrence of the `directional indicator` |dir|.
3. Create |obj| by invoking this algorithm recursively using |path| for |p|.
4. If |dir| is "`!`", emit a new N3 triple (|obj| |pred| `B`) where `B` is a novel blank node. return `B` as the result.
5. Otherwise, |dir| is "`^`", emit a new N3 triple (`B` |pred| |obj|) where `B` is a novel blank node. return `B` as the result.

## William's algorithm:
1. If `path` |p| matches the `pathItem` production, and no prior reduction took place, then return |p| as the result.
2. If `path` |p| matches the `pathItem` production, and some prior reduction took place, then return the last created blank node `B` as the result.
3. Otherwise, separate `p` into two components |pathItem| and |path|, based on the next occurrence of the directional indicator |dir|.
4. If this is the first reduction of `path` |p|, |dir| is "`!`", and |pathItem2| stands for the next item within the |path|, emit a new N3 triple (|pathItem| |pathItem2| `B`) where `B` is a novel blank node.
5. If this is the first reduction of `path` |p|, |dir| is "`^`", and |pathItem2| stands for the next item within the |path|, emit a new N3 triple (`B` |pathItem2| |pathItem|) where `B` is a novel blank node.
6. If this is a subsequent reduction of `path` |p|, |dir| is "`!`", and |B_(n-1)| stands for the previously created blank node, emit a new N3 triple (|B_(n-1)| |pathItem| `B`) where `B` is a novel blank node.
7. If this is a subsequent reduction of `path` |p|, |dir| is "`^`", and |B_(n-1)| stands for the previously created blank node, emit a new N3 triple (`B` |pathItem| |B_(n-1)| where `B` is a novel blank node.

william-vw:
@gkellogg Ahhh .. I understand how your algorithm works now. (This is why I've learned to add disclaimers such as "as far as I understood"!) Honestly it wasn't clear to me until you gave the example. 

I think this is the main thing that tripped me up:
> separate |p| into two components |path| and |pred| from the last occurrence of the directional indicator |dir|.

As I mentioned, |pred| wasn't clear to me; and I thought |path| referred to the last element in the production rule. So IIUC, |path| comprises the resource path with length n-1 (not including the last path item) and |pred| represents the last path item (?) The hint "from the last occurrence of the directional indicator |dir|" was insufficient for me, it seems. I only considered recursion in a "forward" direction, not "backward", hence my comment about appropriateness. (Unsure whether the fixed issue with |obj| would have made a difference in my understanding, really.)

I do appreciate the elegance of the algorithm, but it's a bit unfortunate that it doesn't translate as easily to an event-based parser (i.e., a listener vs. a visitor) since the whole AST is not (yet) available.

I think the following would be clearer to me (no ReSpec formatting at this point):

1. If `path` |path_n| matches the `pathItem` production, then |path_n| can be reduced no further. Return |path_n| as the result.
2. Otherwise, separate |path_n| into two components:

    - |path_(n-1)|: sub-`path` of path_n with length (n-1) that does not include the last `pathItem` in path_n.
    - |pathItem_n|: the last `pathItem` split off from |path_n| at the last occurrence of the `directional indicator` |dir|.

3. Create |result_(n-1)| by invoking this algorithm recursively using |path(n-1)|.
4. If |dir| is "`!`", emit a new N3 triple ( |result_(n-1)| |pathItem_n| `B` ) where `B` is a novel blank node. Return `B` as the result.
5. Otherwise, if |dir| is "`^`", emit a new N3 triple ( `B` |pathItem_n| |result_(n-1)| ) where `B` is a novel blank node. Return `B` as the result.

Perhaps both algorithms can be given - with a toggle "recursive" vs. "iterative" :-)

@TallTed Hope that Gregg's markdown version clarifies the differences better.

TallTed:
The algorithms are definitely much more understandable with the examples.  I think such should be included in the doc!

I don't have a strong feeling for/against either iterative or recursive processing.  If both are viable (which I cannot yet discern), I would suggest they both be included, each with appropriate examples -- optimally with the same example data being handled by both iterative or recursive algorithms -- as it seems likely to me that different implementations/environments would be better served by each.

gkellogg:
> I think the following would be clearer to me (no ReSpec formatting at this point):

Good suggestion.



> Perhaps both algorithms can be given - with a toggle "recursive" vs. "iterative" :-)

I think it would be useful to specify alternates to the algorithm, as that would make it clear that there is _no one right way_ to implement it and only the result matters.

It did occur to me, in the middle of the night, that we've underspecified how these algorithms relate to being in subject, predicate, or object positions, since the necessary resource to use varies. I had it in mind to work more closely with the Turtle _curSubject_ and _curObject_ descriptions, but that looks like pulling on a thread that would have no end.

I'll make an update that includes both algorithms, along with examples of execution for our sample paths, including in at least subject, object and maybe list member contexts, as appropriate.

gkellogg:
@william-vw I took some liberties with your algorithm, which I believe are isomorphic and, IMHO, easier to follow. I also included evaluation examples for each. See what you think.

TallTed:
> decide if we're changing from the use of "Formulae" to "Formulas"; the Team Submission used "Formulae" as the plural of "Formulas".

I don't know if you want(ed) to do this in this PR, but in any case — I would have said Formulae, but [Grammar Monster](https://www.grammar-monster.com/plurals/plural_of_formula.htm#:~:text=Plural%20of%20Formula-,The%20plural%20of%20formula%20is%20formulas%20or%20formulae.,express%20relationships%20between%20various%20quantities.) and [Google's Ngram Viewer](https://books.google.com/ngrams/graph?content=formulas%2Cformulae&year_start=1800&year_end=2000&corpus=15&smoothing=3&share=&direct_url=t1%3B%2Cformulas%3B%2Cc0%3B.t1%3B%2Cformulae%3B%2Cc0) provide strong justification for "Formulas".

gkellogg:
> I don't know if you want(ed) to do this in this PR, but in any case — I would have said Formulae, but [Grammar Monster](https://www.grammar-monster.com/plurals/plural_of_formula.htm#:~:text=Plural%20of%20Formula-,The%20plural%20of%20formula%20is%20formulas%20or%20formulae.,express%20relationships%20between%20various%20quantities.) and [Google's Ngram Viewer](https://books.google.com/ngrams/graph?content=formulas%2Cformulae&year_start=1800&year_end=2000&corpus=15&smoothing=3&share=&direct_url=t1%3B%2Cformulas%3B%2Cc0%3B.t1%3B%2Cformulae%3B%2Cc0) provide strong justification for "Formulas".

We discussed this on a call the other week. It wasn't intentional to move from "Formulae" to "Formulas", although both are correct. "Formulae" is, IMHO, more prosaic, but it is what was used in the original Team Submission, and changing it would seem gratuitous at this point. For example, documentation in my implementation has long used "formulae".

We'll do a more comprehensive update to normalize usage later.

william-vw (commented):
I think the updates are great. I suggested some minor changes on the right-to-left algorithm, and some larger ones on the left-to-right one (see comments for details).

spec/index.html
```diff
@@ -992,50 +992,251 @@ <h3>Path resolution</h3>
         <p><a>Paths</a> are resolved into zero or more <a>N3 triples</a>
           and a single <a>N3 triple element</a> which is used as the
```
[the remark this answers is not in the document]
gkellogg: I think there may besome duplicity in our definitions, which we might straighten out later.

An **N3 triple** is an extension of an **RDF triple**, whereas an **N3 triple element** is the extension of the **RDF resource** concept to include graphs, lists, and variables.

spec/index.html
```diff
@@ -685,34 +968,279 @@ <h3>Escape sequences</h3>
...
+                emit a new <a>N3 triple</a>
+                (<var>B<sub>n</sub></var>
+                 <var>pred<sub>n</sub></var>
+                 <var>obj<sub>n</sub></var>).</li>
+              <li>Return <var>B<sub>n</sub></var> as the result.</li>
+            </ol>
+
+            <aside class="example"
+                   title="Processing :jon!:father right to left">
+              <p>Starting with the <a>path</a> |p| set to `:jon!:father`:</p>
```
gkellogg, line 1051:
```suggestion
              <p>Starting with the <a>path</a> |p| set to `:john!:father`:</p>
```

spec/index.html
```diff
@@ -992,50 +992,251 @@ <h3>Path resolution</h3>
...
+                  </ul>
+                </li>
+                <li>Step 4 matches because <var>dir<sub>n</sub></var> is  "`!`".
+                  The triple (`:john` `:father` <var>B<sub>n</sub></var>) is emitted.</li>
+                <li>Step 6 returns <var>B<sub>n</sub></var>.</li>
+              </ul>
+            </aside>
+
+            <aside class="example"
+                   title="Processing :jon!:father!:mother right to left">
```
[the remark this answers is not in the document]
gkellogg: I considered that. Perhaps the second example should be `:john!:father^:mother`

spec/index.html
```diff
@@ -992,50 +992,251 @@ <h3>Path resolution</h3>
...
+              <p>Starting with the <a>path</a> |p| set to `jon!:father!:mother`:</p>
+              <ul>
+                <li>Step 1 does not apply, as `jon!:father!:mother` is not a
+                  <a href="#grammar-production-pathItem">pathItem</a>.</li>
+                <li>Step 2 separates `jon!:father!:mother` into
+                  <var>p<sub>n-1</sub></var> `:john!:father`,
```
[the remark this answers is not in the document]
gkellogg: I'll do an update to the second examples in each consistently.

spec/index.html
```diff
@@ -992,50 +992,251 @@ <h3>Path resolution</h3>
...
+              MUST be equivalent to using the following algorithm:</p>
+
+            <p>Repeat the following algorithm steps starting by
+              initializing |n| to `0`,
+              separating |p| into two components 
+              <var>B<sub>0</sub></var> and
+              <var>p<sub>0</sub></var>
+              based on the next occurrence of the
+              <a>directional indicator</a> <var>dir<sub>0</sub></var>
+              until return.</p>
```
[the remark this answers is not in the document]
gkellogg: Needs to reference `|p|`, which I'll take care of in a separate commit.

spec/index.html
```diff
@@ -992,50 +992,251 @@ <h3>Path resolution</h3>
...
+
+            <p class="note"><var>B<sub>0</sub></var> starts
+              as the first pathItem in |p|, but is updated to
+              a novel <a>blank node</a> on subsequent iterations.</p>
+
+            <ol>
+              <li>If <var>dir<sub>n</sub></var> is `null`,
+                return <var>B<sub>n</sub></var>.</li>
```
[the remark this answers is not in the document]
gkellogg: Doesn't quite work though, as `n` increments by the number of segments in `|p|`, not the length of the string representation of `|p|`. I'll re-work in a separate commit.

spec/index.html
```diff
@@ -992,50 +992,251 @@ <h3>Path resolution</h3>
...
+              <li>Increment |n|,
+                separate <var>p<sub>n-1</sub></var> into two components
+                <var>item<sub>n</sub></var> and <var>p<sub>n</sub></var>
+                based on the next occurrence of the
+                <a>directional indicator</a> <var>dir<sub>n</sub></var>,
+                if any,
+                and set <var>B<sub>n</sub></var>
+                to a novel <a>blank node</a>.</li>
```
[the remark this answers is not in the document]
gkellogg: We can re-work naming to make less confusing, but there is a difference between `p(n)` and `path(n)`, which are partial paths, and `item(n)` which is a specific pathItem from that path.

spec/index.html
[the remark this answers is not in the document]
william-vw: Yes, indeed, my bad.

spec/index.html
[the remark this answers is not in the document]
william-vw: yes, good point.

spec/index.html
```diff
@@ -1129,102 +1134,97 @@ <h4>Left to Right Algorithm</h4>
...
-                |n| is `0`,
-                <var>B<sub>0</sub></var> is `:john`,
-                <var>p<sub>0</sub></var> is `:father!:mother`, and
-                <var>dir<sub>0</sub></var> is "`!`":</p>
+                `:john^:father!:mother`;
+                |n| is `0` and 
+                <var>B<sub>0</sub></var> is `:john`:</p>
               <ul>
                 <li>Step 1 does not apply,
-                  as <var>dir<sub>0</sub></var> is not `null`.</li>
```
william-vw:
```suggestion
                 as |n| is `0`.</li>
```

spec/index.html
```diff
@@ -1129,102 +1134,97 @@ <h4>Left to Right Algorithm</h4>
...
-                  <var>p<sub>1</sub></var> to `null`, and
-                  <var>B<sub>1</sub></var> a novel <a>blank node</a>.</li>
+                <li>Step 1 does not apply, as |n| is `0`.</li>
+                <li>Step 2 increments |n| to `1`,
+                  sets <var>B<sub>1</sub></var> to a novel <a>blank node</a>,
+                  <var>dir<sub>1</sub></var> to "`!`", and
+                  <var>pred<sub>1</sub></var> to `:father`.</li>
                 <li>Step 3 emits
                   (`:john` `:father` <var>B<sub>1</sub></var>),
                   as <var>dir<sub>0</sub></var> is "`!`"</li>
```
william-vw, line 1186:
```suggestion
                  as <var>dir<sub>1</sub></var> is "`!`"</li>
```

spec/index.html
```diff
@@ -1129,102 +1134,97 @@ <h4>Left to Right Algorithm</h4>
...
-                  as <var>dir<sub>0</sub></var> is "`!`"</li>
-                <li>Looping back to step 1, |n| is 1:
+                  as <var>dir<sub>0</sub></var> is `0`.</li>
+                <li>Step 2 increments |n| to `1`,
+                  sets <var>B<sub>1</sub></var> to a novel <a>blank node</a>,
+                  <var>dir<sub>1</sub></var> to "`^`", and
+                  <var>pred<sub>1</sub></var> to `:father`.</li>
+                <li>Step 4 emits
+                  (<var>B<sub>1</sub></var> `:father` `:john`),
+                  as <var>dir<sub>0</sub></var> is "`^`"</li>
```
william-vw, line 1212:
```suggestion
                  as <var>dir<sub>1</sub></var> is "`^`"</li>
```

william-vw (commented):
some minor suggestions - looks good

spec/index.html
```diff
@@ -685,34 +968,274 @@ <h3>Escape sequences</h3>
...
+
+            <aside class="example"
+                   title="Processing :john^:father!:mother left to right">
+              <p>Starting with the <a>path</a> |p| set to
+                `:john^:father!:mother`;
+                |n| is `0` and 
+                <var>B<sub>0</sub></var> is `:john`:</p>
+              <ul>
+                <li>Step 1 does not apply,
+                  as <var>dir<sub>0</sub></var> is `0`.</li>
```
gkellogg, line 1205:
```suggestion
                  as |n| is `0`.</li>
```

william-vw (approved):

gkellogg:
@william-vw it seems that the way this was merged didn't actually close the PR.

PR #84, opened 2021-05-08
Test updates

william-vw:
Updating test files


gkellogg:
If these changes are going to be made to the test output, then the manifest options have to change to reflect that. By default, the results are expected to include the original dataset and the conclusions. It's through the use of options that things like removing non-inferred triples are specified.

For example, `:cwm_includes_concat` has test options `[test:think true; test:data true;]`, which projects only the facts from the top-level "default" graph, and with `test:think` after repeatedly running rules until stable. To have just the inferred results, it would need to have `test:conclusions true;`.

When I ported these tests over, I didn't change the cwm options specified in the regression manifest, which is why some of them include the data, as well as the inferrences. The various options are described in the [test vocabulary](https://github.com/w3c/N3/blob/master/tests/test.n3) for terms with domain test:Option.

That said, we haven't defined `log:Chaff` and these triples could be removed, as far as I'm concerned. @josd may have a different perspective.

william-vw:
The problem is that this is done very inconsistently - sometimes *some part* of the original dataset is included, sometimes not. Only in rare cases is the entire dataset included in the reference file (e.g., `av.n3`). Not saying this happened during the porting, just saying what I found. 

Even if they were consistent, those options seem rather useless to me since our interest in reasoning tests lies, well, with the reasoning results. I.e., whether the generated inferences include the reference inferences, and vice versa.

william-vw:
I may have commented too soon: the output files I was referring to do seem in line with the `test:data` option (i.e., removing implications from the store), perhaps with a few exceptions.

Still the options seem rather pointless to me .. My test code simply checks whether the inferred statements are included in the ref file, and, vice versa, i.e., whether the ref file statements are included in the inferences (see [N3 code](https://github.com/w3c/N3/blob/master/tests/dogfood/run_reason_tests.n3), lines 42-43). Unsure why more would be needed, aside from e.g., testing the filtering of plain RDF triples (doesn't seem the place for it here). Also the options rather seem geared towards a forward chainer.

gkellogg:
Some of the tests show the value of including the inferred triples along with the original data (e.g., [quant-implies-ref.n3](https://github.com/w3c/N3/pull/84/files#diff-0b6579d29da96e79a867adfd5c8c2cfc34dbbe00283c5e5f3505b50060f81fed); I find value in all the various options used. We should certainly have _some_ set of tests that include the original base data, if not all of the cited formulae.

Generally, when porting over older tests, I tried to stick close to the original expression, but have taken a different tact for newer tests (directories with `cwm_` vs the others. But it's reasonable to simplify the results, as filtering through the inferred triples from the original triples when something goes wrong can take some work. Let's just be sure that the manifest options are consistent with the expected results.

william-vw:
Yes, good point re the quant-implies-ref.n3. I will revise the commit, thanks.

But, if it's not relevant to the test then I would still opt for just listing the inferences (and not performing e.g., filtering of implications). IMO it also makes it much clearer what is expected of the test to the reader.

PR #91, opened 2021-08-09
Semantics

doerthe:
Draft version for definitions and semantics (work in progress)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/N3/pull/91.html" title="Last updated on Sep 23, 2021, 12:11 PM UTC (1d4e490)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/N3/91/fbdd9cd...1d4e490.html" title="Last updated on Sep 23, 2021, 12:11 PM UTC (1d4e490)">Diff</a>

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/N3/pull/91.html" title="Last updated on Jun 29, 2022, 8:59 PM UTC (68bbeb9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/N3/91/fbdd9cd...68bbeb9.html" title="Last updated on Jun 29, 2022, 8:59 PM UTC (68bbeb9)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -894,7 +894,7 @@ Maybe we could get back to that example when we introduce rules or also with bui
 
       For formulas containing nesting we extend the above definition as follows:
       <p>
-        <li>M({F})={F'} where F is isomorphic to F', that is there exists an isomorphosm M' such that M'(F)=F'.</li>
+        <li>M({F})={F'} where F is isomorphic to F'; that is, there exists an isomorphism M' such that M'(F)=F'.</li>
       </l>
       We write G&equiv;G' if G and G' are isomorphic.
       </p>
--- a/spec/index.html
+++ b/spec/index.html
@@ -1009,7 +1009,9 @@ Maybe we could get back to that example when we introduce rules or also with bui
            </li>
         </ul>
        </p>
-
+       <p class="ednote">
+         we need to directly clarify how nested rules are handled, i.e. that we need a new blank node mapping for each nested rule. That is covered in the definition but not really clear.
+       </p>
        <!--
        <p>
 
--- a/spec/index.html
+++ b/spec/index.html
@@ -165,18 +165,26 @@
 
     <section id='language' class=informative>
       <h2>Language</h2>
-      <p>The aim of this section is to provide an informal overview of the N3 language and its different features. Where possible, this section borrows from, or is at least based on, the [[[Turtle]]] specification. More formal definitions will follow in the subsequent sections.
+      <p>The aim of this section is to provide an informal overview of the N3 language and its different features.
+        Where possible, this section borrows from, or is at least based on, the [[[Turtle]]] specification.
+        More formal definitions will follow in the subsequent sections.
+      </p>
+      <p>
+        All examples shown here and in the remainder of this document follow
+        <abbr title="Notation3">N3</abbr>-turtle syntax which is also the standard syntax for <abbr title="Notation3">N3</abbr>.
+        It is however possible, to use alternative syntaxes as long as the translation to <abbr title="Notation3">N3</abbr> turtle syntax is clearly defined.
       </p>
       <section id='n3doc'>
         <h3>N3 document</h3>
-        <p>An <dfn>N3 document</dfn> represents an <a>N3 graph</a> in a compact textual form. An <dfn>N3 graph</dfn>
+        <p>An <dfn>N3 document</dfn> represents an <a>N3 graph</a> in a compact textual form. An <dfn>N3 graph</dfn>, also called <dfn>N3 formula</dfn> or simply
+          <dfn>formula</dfn>
           is a series of <dfn data-lt="n3 triple|statement|triple">N3 statements</dfn>. These are written as <a>triples</a> consisting of:
           <a data-cite="RDF11-CONCEPTS#dfn-subject">subject</a>,
           <a data-cite="RDF11-CONCEPTS#dfn-predicate">predicate</a>, and
           <a data-cite="RDF11-CONCEPTS#dfn-object">object</a>.
           In its simplest form a triple expresses a relation between resources,
           but N3 also contains elements allowing more advanced statements &mdash;
-          <a>cited formulae</a> for quoting sets of statements,
+          <a>quoted graphs</a> for quoting sets of statements,
           <a href="#logimpl">logical implications</a> for If-Then style, rule-based reasoning,
           quantified variables, and so on.
           We introduce these elements below.</p>
@@ -769,13 +777,6 @@ Maybe we could get back to that example when we introduce rules or also with bui
          or <a href="#bnodeprplist">blank node property lists</a> as these can be seen as syntactic sugar.
         </p>
         <p>
-          All examples shown earlier as well as the examples in the remainder of this document follow <abbr title="Notation3">N3</abbr>-turtle syntax which is also the standard syntax for <abbr title="Notation3">N3</abbr>.
-          It is however possible, to use alternative syntaxes as long as the translation to <abbr title="Notation3">N3</abbr> turtle syntax is clearly defined.
-        </p>
-        <p class="ednote">
-          Maybe we put that statement earlier?
-          </p>
-        <p>
         <abbr title="Notation3">N3</abbr> is an extension of RDF.
         We use the following concepts from the [[[RDF11-CONCEPTS]]]: <a data-cite="RDF11-CONCEPTS#section-IRIs">IRI</a>, <a data-cite="RDF11-CONCEPTS#section-triples">RDF triple</a>, <a data-cite="RDF11-CONCEPTS#section-rdf-graph">RDF graph</a>, <a data-cite="RDF11-CONCEPTS#section-triples">subject</a>,
         <a data-cite="RDF11-CONCEPTS#section-triples">predicate</a>, <a data-cite="RDF11-CONCEPTS#section-triples">object</a>, <a data-cite="RDF11-CONCEPTS#dfn-rdf-source">RDF source</a>,
@@ -798,8 +799,8 @@ Maybe we could get back to that example when we introduce rules or also with bui
           <a data-cite="RDF11-CONCEPTS#dfn-blank-node">blank nodes</a>,
           <a data-cite="RDF11-CONCEPTS#dfn-literal">literals</a>,
           <a href="#dfn-universals">universal variables</a>,
-          <a href="#lists">lists</a>,
-          and <a href="#quotes">cited formulae</a>.
+          <!--<a href="#lists">lists</a>,-->
+          and <a href="#quotes">quoted N3 graphs</a>.
           The set of <a data-cite="RDF11-CONCEPTS#dfn-iri">IRIs</a> conatians the implication sign denoted by the <a>built-in</a> <code>log:implies</code>.
           This symbol indicates logical implication and MAY in N3-turtle syntax also be represented using the symbols <code>=&gt;</code>.
           The set of <a data-cite="RDF11-CONCEPTS#dfn-literal">literals</a> contains the two boolean literals <code>false</code> and <code>true</code>.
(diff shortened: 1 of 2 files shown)


spec/index.html
```diff
@@ -839,32 +841,149 @@ <h3>Terms and Formulae</h3>
...
+      </p>
+      <p>
+        <dfn data-lt="semantics">Semantic conditions for closed graphs</dfn>
+        <ul>
+        <li>if E is a literal then I(E) = IL(E)</li>
+        <li>if E is an IRI then I(E) = IS(E)</li>
+        <li>if E is a universal free quoted graph then I(E)=IG(E) and IG(E)&equiv;E</li>
+        <li>if E is a closed triple s p o. with p	&#8800; log:implies, or s&notin;QG&cup;{true,false} or o&notin;QG&cup;{true,false}, then I(E) = true if</li>
+          <ul>
+        <li>I(p) is in IP and the pair <I(s),I(o)> is in IEXT(I(p))</li>
```
pchampin, line 926:
```suggestion
        <li>I(p) is in IP and the pair &lt;I(s),I(o)&gt; is in IEXT(I(p))</li>
```

spec/index.html
```diff
@@ -839,32 +841,149 @@ <h3>Terms and Formulae</h3>
...
+
+            <li>otherwise I(E) = false.</li>
+        </ul>
+        <li>if E is a triple of the form {f} log:implies {g}. then E is true if from I(f)=true follows that I(g)=true. </li>
+        <li>if E is a triple of the form {f} log:implies false. then E is true if I(f)=false. </li>
+          <li>if E is a triple of the form true log:implies {g}. then E is true if I(g)=true. </li>
+          <li>if E is a triple of the form {f} log:implies true. or false log:implies {g}. then I(E)=true.</li>
+          <li> if E=true log:implies false. then I(E)=false., </li>
```
pchampin, line 934: Couldn't we simplify all this by introducing a new function IC (C for "clause"), such that
- If E = { f } is a quoted formula, IC(E) = I(E)
- If E = b a boolean, IC(b) = b
- otherwise, IC(E) is not defined

Then I(C1 => C2) would be true if IC(C1) implies IC(C2)?

---
Also, in these lines, a number of `then E is true` should be `then I(E)=true`

spec/index.html
```diff
@@ -839,32 +841,149 @@ <h3>Terms and Formulae</h3>
...
+       <p>
+        We call an <a>N3 formula</a> <dfn id="uv-free">universal free</dfn> if its direct and nested
+        components do not contain <a>universal variables</a>. We denote the set of universal free quoted graphs by QG.</p>
+
+      <p> We call an <a>N3 formula</a> <dfn data-lt="closed">closed</dfn> if its direct and nested components do not contain <a>universal variables</a>,
+        its direct components do not contain <a>blank nodes</a> and its nested components do only contain blank nodes if they are not subject or object to the predicate
+        <code>log:implies</code>.
+      </p>
+      <p class="ednote">
+        Clarify what "contains" means.
```
pchampin, line 874: May be this note dates back to an older version of the text, but I find the term "contains" in this section clear enough. Everytime it is used, its is about a set, and so I interpret it as refering to ∈(element of).

domel:
My general comment is that things like `log:implies` should be inside `<code>log:implies</code>`. Another general suggestion is that `<math>` is used incorrectly .

spec/index.html
```diff
@@ -874,20 +878,25 @@ <h3>Terms and Formulae</h3>
...
-        <li>M({F})={F'} where the nested graph F is isomorphic to F'.</li>
-      </ol>
-      We write G&equiv;G'.
+      </l>
+
+      </p>
+
+      For formulas containing nesting we extend the above definition as follows:
+      <p>
+        <li>M({F})={F'} where F is isomorphic to F', that is there exists an isomorphosm M' such that M'(F)=F'.</li>
```
TallTed, line 897:
```suggestion
        <li>M({F})={F'} where F is isomorphic to F'; that is, there exists an isomorphism M' such that M'(F)=F'.</li>
```

spec/index.html
```diff
@@ -874,20 +878,25 @@ <h3>Terms and Formulae</h3>
...
+      </l>
+      We write G&equiv;G' if G and G' are isomorphic.
```
TallTed, line 899:
```suggestion
      We write G&equiv;G', if G and G' are isomorphic.
```

spec/index.html
```diff
@@ -983,6 +995,54 @@ <h4>Blank nodes</h4>
...
       <section>
         <h4>Universals</h4>
+        <p class="ednote">
+          Definitions below are not finished yet.
+        </p>
+        Universals are treated as indicating universal statements about things. They are typically used in rules. Universal variables are quantified
+        on top of the graph.
+
+        <p>
+        For the interpretation of universal variables we define two kinds of mappings.
```
TallTed, line 1058:
```suggestion
        For the interpretation of universal variables, we define two kinds of mappings.
```

spec/index.html
```diff
@@ -983,6 +995,54 @@ <h4>Blank nodes</h4>
...
+        <ul>
+          <li>A mapping U from the set of universal variables to <math>IP	&cup; IR</math></li>
+          <li>A mapping V from the set of universal variables to the union of literals, iris, blank nodes and closed graphs. </li>
```
TallTed, line 1061:
```suggestion
          <li>A mapping V from the set of universal variables to the union of literals, iris, blank nodes, and closed graphs. </li>
```

spec/index.html
```diff
@@ -983,6 +995,54 @@ <h4>Blank nodes</h4>
...
+        </ul>
+        Suppose I is an inperpretation covering blank nodes. Define the mappings [I+U] and [I+V] to be I on names, and A on blank nodes on the set:
+        [I+A](x)=I(x) when x is a name and [I+A](x)=A(x)
+        when x is a blank node;
+        and extend this mapping to triples and N3 graphs using the rules given above for closed graphs.
+        Then the semantic conditions for an N3 graph are:
+        </p>
+        <p>
+          Suppose I is a simple interpretation and A is a mapping from a set of blank nodes to the
```
TallTed, line 1070:
```suggestion
          Suppose I is a simple interpretation, and A is a mapping from a set of blank nodes to the
```

spec/index.html
```diff
@@ -983,6 +995,54 @@ <h4>Blank nodes</h4>
...
+          <ul>
+            <li>
+               there exists some A such that for all triples T in E for which p	&#8800; log:implies, or s&notin;QG&cup;{true,false} or o&notin;QG&cup;{true,false}
+               [I+A](T)=true.
+            </li>
+            <li> for each triple in E of the form {f} log:implies {g}. there exist some blank node mappings A
+              and A' such that from [I+A](f)=true follows that [I+A'](g)=true. </li>
+            <li>if E is a triple of the form {f} log:implies false. then E is true if [I+A](f)=false for all blank node mappings A. </li>
+              <li>if E is a triple of the form true log:implies {g}. then E is true if [I+A](g)=true for some A. </li>
+              <li>if E is a triple of the form {f} log:implies true. or false log:implies {g}. then I(E)=true.</li>
```
TallTed, line 1090:
```suggestion
            <li> for each triple in E of the form {f} log:implies {g}., there exist some blank node mappings A
              and A' such that from [I+A](f)=true, it follows that [I+A'](g)=true. </li>
            <li>if E is a triple of the form {f} log:implies false., then E is true if [I+A](f)=false for all blank node mappings A. </li>
              <li>if E is a triple of the form true log:implies {g}. and [I+A](g)=true for some A, then E is true. </li>
              <li>if E is a triple of the form {f} log:implies true. or false log:implies {g}., then I(E)=true.</li>
```

spec/index.html
doerthe: thank you for all comments, I just discovered them and will take a look.

doerthe:
> My general comment is that things like `log:implies` should be inside `<code>log:implies</code>`. Another general suggestion is that `<math>` is used incorrectly .

You are right about both and I was wondering how to best do it. So, what is your suggestion (I will simply follow :) ): all N3 code example in code context? Other context you'd prefer? I thought I use math in all cases where I have formulas (or something like a formula), what would you use here?

spec/index.html
doerthe: You are right that the handling of the implication should get less ugly, I will fix it as suggested.

The mistake with I(E) is already corrected.

domel:
> > My general comment is that things like `log:implies` should be inside `<code>log:implies</code>`. Another general suggestion is that `<math>` is used incorrectly .
> 
> You are right about both and I was wondering how to best do it. So, what is your suggestion (I will simply follow :) ): all N3 code example in code context? Other context you'd prefer? I thought I use math in all cases where I have formulas (or something like a formula), what would you use here?

I think that the best option is to use `<code>` in the regular text. In math block `<math>` is OK but you need to use other MathML elements inside. If you don't know MathML, I suggest using LibreOffice Math and then copying the formula to the specification.

spec/index.html
```diff
@@ -165,18 +165,26 @@ <h3>Namespaces</h3>
...
+      <p>
+        All examples shown here and in the remainder of this document follow
+        <abbr title="Notation3">N3</abbr>-turtle syntax which is also the standard syntax for <abbr title="Notation3">N3</abbr>.
+        It is however possible, to use alternative syntaxes as long as the translation to <abbr title="Notation3">N3</abbr> turtle syntax is clearly defined.
       </p>
       <section id='n3doc'>
         <h3>N3 document</h3>
-        <p>An <dfn>N3 document</dfn> represents an <a>N3 graph</a> in a compact textual form. An <dfn>N3 graph</dfn>
+        <p>An <dfn>N3 document</dfn> represents an <a>N3 graph</a> in a compact textual form. An <dfn>N3 graph</dfn>, also called <dfn>N3 formula</dfn> or simply
+          <dfn>formula</dfn>
```
TallTed, line 180:
```suggestion
          <dfn>formula</dfn>,
```

spec/index.html
```diff
@@ -809,23 +810,27 @@ <h3>Terms and Formulae</h3>
...
           we indicate lists by using round brackets <code>()</code>.
           The empty list in <abbr title="Notation3">N3</abbr> is equivalent to the empty list in RDF and can
           thus also be represented using <a href="https://www.w3.org/TR/rdf11-mt/#rdf-interpretations">rdf:nil</a>.
         </p>
+      -->
         <p>
-          <dfn id="def-quotes">Cited formulae</dfn> (sometimes also called quoted graphs) are <a>N3 formulae</a> which are surrounded by quoting signs.
-          In N3-turtle syntax, we denote these quoting signs as curly brackets <code>{}</code>.
+          <dfn id="def-quotes">Quoted Graphs</dfn> (sometimes also called cited formulae) are <a>N3 formulae</a> which are surrounded by quoting signs.
+          In N3-turtle syntax, we denote these quoting signs as curly brackets <code>{}</code>. We denote the set of quoted graphs by <math>QG</math>.
```
TallTed, line 823:
```suggestion
          In N3-turtle syntax, we denote these quoting signs as curly brackets, also known as braces: <code>{}</code>. We denote the set of quoted graphs by <math>QG</math>.
```

spec/index.html
```diff
@@ -839,33 +844,261 @@ <h3>Terms and Formulae</h3>
...
 
 
       <section id='sinterpret'>
         <h3>Simple N3 interpretations</h3>
+      <p>
+      A <dfn data-lt="simpleI">simple N3 interpretation</dfn> is a structure I consisting of:
+      <ol>
+        <li> A non-empty set IR of resources, called the domain or universe of I.</li>
+        <li> A set IP, called the set of properties of I.</li>
+        <li> A mapping IEXT from IP into the powerset of IR x IR i.e. the set of sets of pairs < x, y > with x and y in IR.</li>
```
TallTed, line 915:
```suggestion
        <li> A mapping IEXT from IP into the powerset of IR x IR, i.e., the set of sets of pairs < x, y > with x and y in IR.</li>
```

spec/index.html
```diff
@@ -839,33 +844,261 @@ <h3>Terms and Formulae</h3>
...
+             <li>E does not contain the rule true log:implies false.;</li>
+          <!--   <li>for each triple T in E of the form {f} log:implies true. or false log:implies {g}. then I(E)=true.</li>
+             <li> if E is of the  form false log:implies false., true log:implies true., or false log:implies true., then I(E)=true.</li> -->
+           <li>
+             otherwise I(E)= false.
+           </li>
+        </ul>
+       </p>
+       <p class="ednote">
+         we need to directly clarify how nested rules are handled, i.e. that we need a new blank node mapping for each nested rule.
```
TallTed, line 1013:
```suggestion
         we need to directly clarify how nested rules are handled, i.e., that we need a new blank node mapping for each nested rule.
```

PR #103, opened 2022-06-27
Fix some typos in the time vocabulary.

gkellogg:


gkellogg:
Looks like there's some cruft in the spec which ReSpec is just now picking up. I'll address that in the iriPropertList PR.

PR #105, opened 2022-09-10
Updated NS files

william-vw:
I updated the `rdfs:comment`s in NS files for consistency and removed the schema part (as this part is now separately generated from high-level domain schemas). There's still plenty of work to be done - e.g., not all built-ins have an `rdfs:comment`s, and many of them are incomplete (e.g., not describing that the object can be calculated from the subject). 

I didn't regenerate the HTML files as this is done differently in my current workflow. Future work will be to integrate all this into a single process.

### DIFF
--- a/ns/list.n3
+++ b/ns/list.n3
@@ -21,176 +21,94 @@
 list:append a rdf:Property, owl:FunctionalProperty;
     rdfs:label "append";
     rdfs:domain log:List;
-    rdfs:comment """
-Concatenates separate lists into a single list
+    rdfs:comment """Concatenates separate lists into a single list. 
 
-**schema**:
-`($a_1 .. $a_n) list:append $a_s`
+`true` if and only if the subject is a list of lists and the concatenation of all those lists is `$o`.
 
-**definition**:
-`true` if and only if the subject is a list of lists and the concatenation of all those lists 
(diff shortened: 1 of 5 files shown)


gkellogg (approved):
The updates to the namespace files, themselves, seem fine. I'll have to reserve judgement on the updated way to generate the documentation.

domel:
Perhaps it is worth separating the examples from `rdf:comment`? I suggest using [skos:example](https://www.w3.org/TR/skos-reference/#examplel).

domel:
Maybe also other SKOS properties, e.g. [skos:definition](https://www.w3.org/TR/skos-reference/#definition).

william-vw:
> The updates to the namespace files, themselves, seem fine. I'll have to reserve judgement on the updated way to generate the documentation.

Thanks Gregg. Re documentation generation, the major update is generating a string from the high-level domain schemas (i.e., the "schema" part) - the rest would be very similar.

william-vw:
> Maybe also other SKOS properties, e.g. [skos:definition](https://www.w3.org/TR/skos-reference/#definition).

@domel Good idea. Would you be willing to take the lead on that?

josd (approved):

PR #122, opened 2023-01-17
Feature core builtin specs

tduval-unifylogic:
Requesting to merge into master


tduval-unifylogic:
That's my bad...should be added to .gitignore...correct!

PR #146, opened 2023-02-26
fixes #140 "Proposed editorial changes" only.

### LINKED ISSUE #140: Named Graphs extension vs other ways
https://w3c.github.io/N3/spec/#rel-n3-trig explains:
- `graph :g {:s :p :o}` is TriG not N3
- `:g sem:quotedGraph {:s :p :o}` is N3

I have two questions/desires:
- What would it take to support the TriG notation in N3? Is it impossible?
- Would it be possible to do it with the `id` keyword as used in https://w3c.github.io/N3/spec/#iriprplist :
```
{id :g  
  :s1 :p1 :o1.
  :s2 :p2 :o2}
```

---
Proposed editorial changes for that section
- rename from ` TriG` to ` TriG vs N3 Named Graphs`
- emphasize that the prefix 
  `PREFIX sem: <https://example.org/ns/semantics#>`
  is correct, as can be seen in https://github.com/IDLabResearch/N3NamedGraphSemantics/blob/master/3_3_quote/example.n3#L2.
  I was very surprised by this prefix, so needed reassurance.

VladimirAlexiev:


spec/index.html
```diff
@@ -1875,10 +1875,10 @@ <h3>SPARQL</h3>
         [[[SPARQL11-QUERY]]].</p>
     </section>
     <section id="rel-n3-trig">
-      <h3>TriG</h3>
+      <h3>TriG vs N3 Named Graphs</h3>
```
william-vw, line 1878: This addition does not make much sense to me, as each section pertains to the similarity of N3 with another language (or aspect thereof). In this vein, adding "Named Graphs" as a subsection could work.

william-vw (changes requested):

PR #162, opened 2023-03-10
Update manifest.ttl

jeswr:
It is preferable for the manifest to have exactly the same URI as the document as is the case for the other manifests in this repo.


jeswr:
@william-vw

PR #184, opened 2023-06-04
Fixed some mispelling and line breaks

phochste:
Fixing some misspelled words and added some line breaks to make the example better readable.


domel (approved):
A request for the future is if you could divide PRs by subject, it would be easier to check.

william-vw (approved):

PR #188, opened 2023-06-14
Fix typo: `:weater` → `:weather`

Peeja:
Just a little 🧹

***
<a href="https://pr-preview.s3.amazonaws.com/Peeja/N3/pull/188.html" title="Last updated on Jun 14, 2023, 2:45 PM UTC (7b86e9a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/N3/188/5fa35bf...Peeja:7b86e9a.html" title="Last updated on Jun 14, 2023, 2:45 PM UTC (7b86e9a)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -933,7 +933,7 @@ _:a ex:town [ ex:name "Springfield" ] . -->
       <p>
         Hence, when the <a>premise</a> holds, then we can safely infer the <a>conclusion</a>.
         This is also referred to as <dfn data-lt="fire|firing">firing</dfn> the rule.
-        For the example above, if we state `:weather a :Raining`, then we can safely infer `:weater a :Cloudy`.
+        For the example above, if we state `:weather a :Raining`, then we can safely infer `:weather a :Cloudy`.
       </p>
 
       <p class="note">


william-vw:
thanks for this (and sorry for my late answer - I am currently on holiday)

spec/index.html
```diff
@@ -933,7 +933,7 @@ <h3>N3 Rules</h3>
       <p>
         Hence, when the <a>premise</a> holds, then we can safely infer the <a>conclusion</a>.
         This is also referred to as <dfn data-lt="fire|firing">firing</dfn> the rule.
-        For the example above, if we state `:weather a :Raining`, then we can safely infer `:weater a :Cloudy`.
+        For the example above, if we state `:weather a :Raining`, then we can safely infer `:weather a :Cloudy`.
```
TallTed, line 936: Except that [sunshowers](http://dbpedia.org/resource/Sunshower) — rain from blue skies — do happen, so this is not a 100% safe inference. Maybe 99%, maybe more, maybe less. I'm not a meteorologist.

william-vw: @TallTed Sunshowers still involve clouds - don't know where else the rain would come from - but (possibly) not right above where it is raining.

If we will start reviewing our examples by their realism, I think we're in for quite a ride. (Although it can be argued they're more or less accurate in an imaginary MCU.)

TallTed: I think "can safely infer" remains too strong. I think "might infer" allows for there to be other criteria in play, as well as this being a possibly erroneous inference.
```suggestion
        For the example above, if we state `:weather a :Raining`, then we might infer `:weather a :Cloudy`.
```

Peeja: @TallTed Except any time it's rainy, it really *is* cloudy. We know that, because it says it in the document! Whether that rule holds in reality under the intuitive interpretation is a different question, and really only confuses the spec.

If anything, I'd add a remark where the rule is introduced pointing out that real life isn't that simple, but I don't think it's really necessary. At this point in the doc, though, it's absolutely correct to "safely infer" that `:weather a :Cloudy` from the statements given.

william-vw: > We know that, because it says it in the document! Whether that rule holds in reality under the intuitive interpretation is a different question, and really only confuses the spec.

@Peeja Good point! We don't want to get into the weeds with probabilistic logic...

TallTed: Invisible clouds (i.e., where the mist is dispersed enough to be unresolvable by the human eye) do not make "cloudy" weather/skies.

"We know that, because it says it in the document!" This falls back to the original major failure of RDF — that anything encoded in RDF is truth, now and forever. I can say anything in a document, true or false, and readers simply accepting and treating it as truth will lead to catastrophic failures.

All the above taken in... 

I suggest we use a different example than `:weather a :Raining`, and only suggest an inference which *really is* safe.

william-vw: > "We know that, because it says it in the document!" This falls back to the original major failure of RDF — that anything encoded in RDF is truth, now and forever. I can say anything in a document, true or false, and readers simply accepting and treating it as truth will lead to catastrophic failures.

This opens up a whole can of worms - if we want to go for realism then we should drop the MCU examples (e.g., webslingers can fly...) as well. I'd rather this not get turned into a discussion on the underlying RDF semantics - this is not a problem (if it is) that will be solved by N3.

Issue #185, opened 2023-06-05
Is `:a :b {}` the same triple as `:a :b true`


jeswr:
In https://github.com/rdfjs/types/issues/37 there has been a discussion around whether parsers should interpret `{}` as the literal `true`.

To me this seems reasonable. But I think an explicit decision needs to be made on this at a spec level, rather than it being implicit dependent on ones' interpretation of whether the last statement in [ground graph semantic](https://w3c.github.io/N3/spec/semantics.html#ground-graphs) is a syntactic or semantic one.

Once a decision is made I am happy to follow up with a PR containing either the positive or negative test case.

doerthe:
The whole discussion here has several aspects.

First of all: In N3 and also in RDF the empty graph (already in N3's base interpretation and RDF's simple entailment) is always true. 

But in the examples you give, we do not just have the empty graph, we do have the empty graph used as a term and that is different.  If we have an N3 implication like

` {}=>{:birds :can :fly}.` 

We can translate the premise to an FOL-like truth, all together 

⊤ → can(birds,fly)

But: true is a boolean literal and not a formula, so the question here is rather whether we want to translate ⊤ to true. To add another example:

`:doerthe :denies {}.`

(freely translated to "Dörthe denies the truth" :)) then we have in FOL-style something like

denies(doerthe, ⊤)

Again, here the ⊤ stands for true as a formula and it is up to us whether we think that this is the same as the boolean literal true.

Currently we see them as equal, we state that true is syntactic sugar for {}. So, following your request, I would either make more explicit that we do that or define some IRI n3:trueGraphTerm instead. 

Could you (or others) think of any use cases in which we would like to be able to distinguish between true and {}? So far, I think it is a good solution to treat {} and true as equal, but I could be wrong.

jeswr:
@william-vw ?

william-vw:
> Could you (or others) think of any use cases in which we would like to be able to distinguish between true and {}? So far, I think it is a good solution to treat {} and true as equal, but I could be wrong.

Not really, no. Inversely, I can imagine use cases where one would want to differentiate between `false` and `{ }`, e.g., when a builtin fails (`false`) versus returning an empty formula (nothing was found / calculated). I can't think of an example where we'd have to distinguish between a `true` vs. `{ }` value, but maybe I'm lacking the imagination...

doerthe:
I thought about it and think that it might be problematic that `{}` and `true` might get two raw types. 
[...]

Issue #157, opened 2023-03-09
Evaluation test cases for `<=`


jeswr:
Since EYE now interprets `<=` as `log:isImpliedBy` (and I assume this plans to be standardised) I would greatly appreciate if this could be reflected in some test cases.

william-vw:
@jeswr Feel free to add a PR with a suitable test case.

PR #202, opened 2023-07-31
feat: add eval test cases for empty graph and isImpliedBy

jeswr:
Test cases addressing:
 - https://github.com/w3c/N3/issues/185
 - https://github.com/w3c/N3/issues/157
 
There may be a problem with some of these test cases as some:
 - Contain literals in the subject position
 - Contain variables
 
 Neither of which is actually permitted in the n-quads spec. A solution for those test cases could be to have the same content for the result but contained in another n3 document. What do you think @william-vw ?


william-vw:
@jeswr 
> A solution for those test cases could be to have the same content for the result but contained in another n3 document.

Just to be sure, you mean keeping the results in an n3 results file instead of n-triples / n-quads?

jeswr:
> Just to be sure, you mean keeping the results in an n3 results file instead of n-triples / n-quads?

Correct

william-vw:
@jeswr yes I think that makes a lot of sense!

tests/N3Tests/isImpliedBy/isImpliedBy_bcRule.nq
```diff
@@ -0,0 +1,6 @@
+?a <http://example.org/b1> <http://example.org/c1> _:b0 .
+?a <http://example.org/b> <http://example.org/c> _:b1 .
+_:b0 <http://www.w3.org/2000/10/swap/log#isImpliedBy> _:b1 _:b2 .
```
jeswr, line 3: > @jeswr yes I think that makes a lot of sense!

So it turns out that doesn't work on files like this because N3 doesn't fully support `n-quads` syntax, in particular having a graph term like this one.

Do you think it is worth opening up discussion as to whether this syntax should be supported in the spec @william-vw @josd .

josd: Yes @jeswr this makes a lot sense and the latest EYE does an experimental parsing of n-quads.
```
?a <http://example.org/b1> <http://example.org/c1> _:b0 .
?a <http://example.org/b> <http://example.org/c> _:b1 .
_:b0 <http://www.w3.org/2000/10/swap/log#isImpliedBy> _:b1 _:b2 .
```
is now `--pass`'ed as
```
<<_:sk_0 <http://example.org/b1> <http://example.org/c1>>> <http://www.w3.org/2000/10/swap/graph#name> _:e_b0_1.
<<_:sk_1 <http://example.org/b> <http://example.org/c>>> <http://www.w3.org/2000/10/swap/graph#name> _:e_b1_1.
<<_:e_b0_1 <= _:e_b1_1>> <http://www.w3.org/2000/10/swap/graph#name> _:e_b2_1.
```

jeswr: Generally the results of eval tests are given in an n-quads format (in particular see the [rdf-tests repo](https://github.com/w3c/rdf-tests/tree/main/rdf/rdf11)) as it is straightforward to parse and as such is a good syntax in which to unambiguously provide the baseline result.

In terms of what I'm proposing, it is that the following:

```n3
@prefix : <http://example.org/> .
"a" :b { :c :d :e } .
```

Be considered equivalent to:

```n3
@prefix : <http://example.org/> .
"a" :b _:b0 .
:c :d :e _:b0 .
```

Parsers over in the RDF/JS world like [N3.js](https://github.com/rdfjs/N3.js/) already interpret the first statement as the following RDF/JS quads:


```ts
const result = [
quad(
  literal("a"),
  namedNode("http://example.org/b"),
  blankNode("b0"),
),
quad(
  namedNode("http://example.org/c"),
  namedNode("http://example.org/d"),
  namedNode("http://example.org/e"),
  blankNode("b0"),
)]
```

josd: > In terms of what I'm proposing, it is that in N3 the following:
> 
> ```
> @prefix : <http://example.org/> .
> "a" :b { :c :d :e } .
> ```
> 
> Be considered equivalent to:
> 
> ```
> @prefix : <http://example.org/> .
> "a" :b _:b0 .
> :c :d :e _:b0 .
> ```

IMHO this is not the same!

josd: > In terms of what I'm proposing, it is that in N3 the following:
> 
> ```
> @prefix : <http://example.org/> .
> "a" :b { :c :d :e } .
> ```
> 
> Be considered equivalent to:
> 
> ```
> @prefix : <http://example.org/> .
> "a" :b _:b0 .
> :c :d :e _:b0 .
> ```

They are different, the former is a graph term, the latter is a graph by reference.
I have added initial support for n-quads in eye and this example round trips fine:
```
$ cat nq2.n3
@prefix : <http://example.org/> .

"a" :b _:b0 .
:c :d :e _:b0 .
```
```
$ eye --quiet --nope nq2.n3 --pass
@prefix : <http://example.org/>.

"a" :b _:e_b0_1.
:c :d :e _:e_b0_1.
```
One can of course compose graph term with the following rule in
```
$ cat nq3.n3
@prefix log: <http://www.w3.org/2000/10/swap/log#>.
@prefix graph: <http://www.w3.org/2000/10/swap/graph#>.
@prefix : <http://example.org/> .

"a" :b _:b0 .
:c :d :e _:b0 .
:f :g :h _:b0 .

{   ?A ?B ?C.
    ({?D ?E ?F} {?D ?E ?F ?C} ?L) log:collectAllIn ?SCOPE.
    ?G graph:list ?L.
} => {
    ?A ?B ?G.
}.
```
```
$ eye --quiet --nope nq3.n3 --pass
@prefix log: <http://www.w3.org/2000/10/swap/log#>.
@prefix graph: <http://www.w3.org/2000/10/swap/graph#>.
@prefix : <http://example.org/>.

"a" :b _:e_b0_1.
"a" :b {
    :c :d :e.
    :f :g :h.
}.
:c :d :e _:e_b0_1.
:f :g :h _:e_b0_1.
:c :d :e {
    :c :d :e.
    :f :g :h.
}.
:f :g :h {
    :c :d :e.
    :f :g :h.
}.
```

william-vw: > Generally the results of eval tests are given in an n-quads format (in particular see the [rdf-tests repo](https://github.com/w3c/rdf-tests/tree/main/rdf/rdf11)) as it is straightforward to parse and as such is a good syntax in which to unambiguously provide the expected result of parsing.

Why would n-quads be better to provide expected results - what would the graph label indicate? I'm not finding n-quads result files in the folder you reference (aside from the rdf-n-quads folder, but those are not result files). Not trying to be difficult here, I am just wondering what the added value would be for an individual results file. 

It's a much bigger discussion, for sure, but I feel that graph terms can be used to group triples belonging to the same graph (these could be made "referentially transparent" as well).

jeswr:
I've updated the files @william-vw .

I'll also respond to https://github.com/w3c/N3/pull/202#discussion_r1282267906 in due course when I have time to write a full response.

tests/N3Tests/manifest-parser.ttl
```diff
@@ -1050,15 +1050,15 @@
 :isImpliedBy_graphs
         a              test:TestN3Eval ;
         mf:action      <isImpliedBy/isImpliedBy_graphs.n3> ;
-        mf:result      <isImpliedBy/isImpliedBy_graphs.nq> ;
+        mf:result      <isImpliedBy/isImpliedBy_graphs_result.n3> ;
         mf:name        "isImpliedBy Eval";
         rdfs:comment   "<= should be parsed as log:isImpliedBy in a complex use case";
```
william-vw, line 1055: unsure what is meant by complex use case?

william-vw (approved):
looks good (aside from minor comment)

tests/N3Tests/manifest-parser.ttl
```diff
@@ -1023,6 +1023,46 @@
...
+        mf:name        "isImpliedBy Eval";
+        rdfs:comment   "<= should be parsed as log:isImpliedBy";
+        rdft:approval  rdft:Approved .
+
+:isImpliedBy_graphs
+        a              test:TestN3Eval ;
+        mf:action      <isImpliedBy/isImpliedBy_graphs.n3> ;
+        mf:result      <isImpliedBy/isImpliedBy_graphs_result.n3> ;
+        mf:name        "isImpliedBy Eval";
+        rdfs:comment   "<= should be parsed as log:isImpliedBy in a complex use case";
```
jeswr, line 1055:
```suggestion
        rdfs:comment   "<= should be parsed as log:isImpliedBy with graph as premise and object";
```

tests/N3Tests/manifest-parser.ttl
```diff
@@ -1023,6 +1023,46 @@
...
+        mf:name        "isImpliedBy Eval";
+        rdfs:comment   "<= should be parsed as log:isImpliedBy";
+        rdft:approval  rdft:Approved .
+
+:isImpliedBy_graphs
+        a              test:TestN3Eval ;
+        mf:action      <isImpliedBy/isImpliedBy_graphs.n3> ;
+        mf:result      <isImpliedBy/isImpliedBy_graphs_result.n3> ;
+        mf:name        "isImpliedBy Eval";
+        rdfs:comment   "<= should be parsed as log:isImpliedBy with graph as premise and object";
```
jeswr, line 1055:
```suggestion
        rdfs:comment   "<= should be parsed as log:isImpliedBy with N3 graphs as subject and object";
```

tests/N3Tests/manifest-parser.ttl
jeswr: Not sure either - resolved in https://github.com/w3c/N3/pull/202/commits/59f2ae74df01baa1ee2c3bad901ce578c644c252.

I don't have permissions to merge on this repo so will leave that to you.

william-vw (approved):

PR #216, opened 2024-05-12
Remove unbalanced bracket

eyusupov:
Looks like it was missed when applying the convention for omitting the inner graph parentheses.


doerthe:
You were right there (and I thought I got all brackets :) )

PR #218, opened 2024-05-12
Fix typo

eyusupov:


spec/semantics.html
```diff
@@ -1181,7 +1181,7 @@ <h2>Constraints for `log:implies`</h2>
 \qquad        &lt;\{\},\{\},\{(\text{:birds},\text{:can}, \text{:fly} )\}&gt; \\
           )\})
         $</div>
-        <div>&ndash; in concrete syntax: <code>{:simon :says {:birds :can :fly}}=>{:bird :can :fly}. </code></div>
+        <div>&ndash; in concrete syntax: <code>{:simon :says {:birds :can :fly}}=>{:birds :can :fly}. </code></div>
```
TallTed, line 1184: It would really be better to change all instances of `:birds` to `:bird`...

PR #219, opened 2024-05-15
Use single comment style

eyusupov:
Use the comment style as specified in [https://www.w3.org/TR/REC-xml/#sec-notation](XML spec), to which the Turtle spec links. Currently both `/* .. */` and `#` are used


TallTed:
Correcting and clarifying initial comment --

> Use the comment style as specified in [XML spec](https://www.w3.org/TR/REC-xml/#sec-notation), i.e., `/* ... */`

PR #221, opened 2025-09-09
Fix wording in log:equalTo example

eyusupov:
Example 64 in log:equalTo builtin description has the subject list and object list mixed up.


doerthe:
I totally forgot about. that. So yes, we need to find the source.

william-vw:
already done -
https://github.com/w3c-cg/n3Builtins/blob/main/spec/src/log/equalTo.n3#L158

doerthe:
As always: one step ahead :)

PR #227, opened 2026-04-23
Fixing broken layout 

phochste:
The current rendering of the semantics specs shows literal HTML fragments at several places instead of their rendered version.

Apperently when removing some newlines in the HTML fixed one by one the issue with the broken layout. 

But, also removing the `markdown` setting in the Respec settings (without touching the HTML) did the same trick.

### DIFF
--- a/spec/semantics.html
+++ b/spec/semantics.html
@@ -48,7 +48,6 @@
         shortName: "n3-semantics",
         xref: "web-platform",
 	      group: "n3-dev",
-        format: "markdown",
         edDraftURI: null,
         latestVersion: "https://w3c.github.io/N3/reports/20230703/semantics.html"
       };


TallTed:
> But, also removing the `markdown` setting in the Respec settings (without touching the HTML) did the same trick.

But it does nothing good for the [many instances of ``` ` ``` Markdown sugar](https://w3c.github.io/N3/spec/) for HTML `<code>` tags! I have not reviewed the current editor's draft for other Markdown tags.

This PR must be backed out. It should not have been merged so quickly.

There should be a proper issue raised for the "broken layout" of "literal HTML fragments at several places instead of their rendered version", and probably the PR to resolve the new issue will need to "remove some newlines in the HTML [to fix] one by one the issue with the broken layout."

Quick and dirty fixes are rarely the fix they may appear to be at first glance.

phochste:
@TallTed Do you know how to load a specific version of respec in a HTML page? The markdown option suddenly began misbehaving ..without any changes in the HTML.

I't can't be that adding a new line at the end of a `<p>..</p>` statement suddenly breaks the rendering of the HTML code. Would like to see if this is also the case when loading a previous version of respec.

phochste:
I checked again the rendering of Markdown sugar backticks  code. Looks correct for me. it could be that respec already does markdown rendering of such things in the newest version and the extra `markdown` in the configuration messed it up by additional markdown rendering that was already done.

william-vw:
@phochste if this PR left the webpages broken, can you please recommend another PR for fixing it?

PR #228, opened 2026-04-24
Fix semantics layout. Remove 'format' property from semantics object

phochste:
Apparently the 'markdown' setting also needs to be removed from the reports version.


TallTed:
> Apparently the 'markdown' setting also needs to be removed from the reports version.

Don't do it. See [my comment on the other related PR](https://github.com/w3c/N3/pull/227#issuecomment-4335153403).

william-vw:
@phochste Would you like to amend the PR?

PR #229, opened 2026-05-05
Fixing the semantics layout

phochste:
Adding a fix for the semantics spec layout issues.

### DIFF
--- a/reports/20230703/semantics.html
+++ b/reports/20230703/semantics.html
@@ -48,7 +48,6 @@
         shortName: "n3-semantics",
         xref: "web-platform",
 	      group: "n3-dev",
-        format: "markdown",
         edDraftURI: null,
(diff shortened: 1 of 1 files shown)


william-vw:
@phochste Thanks. Can you please confirm that removing the format option will not lead to issues as raised by @TallTed in https://github.com/w3c/N3/pull/227#issuecomment-4315298605?

phochste:
@william-vw I can not confirm that removing the format option will stop the rendering of markdown directives in the spec file. The redering of backticks (`) , dollar ($) still provides inline code rendering and mathematical rendering.

william-vw:
>  I can not confirm that removing the format option will stop the rendering of markdown directives in the spec file. 

That was not the question :-)

phochste:
My bad writing. The fix of removing with format will not lead to the issues raised by @TallTed. The rendering looks ok.

